PR #18, opened 2020-11-21
Add option to toggle pushing to remote

weph:
I have a vault that I want to backup but don't want to push to a remote repository. Actually, this already works in the current version but the error notification shows up every n minutes and it gets a bit annoying.

I could have worked around this by creating a local bare repository in another directory but then I thought this might be a nice little addition to your plugin. Especially for people who are not so much into git and don't know about bare repos and stuff.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -32,7 +32,7 @@ export default class ObsidianGit extends Plugin {
     async onload() {
         let statusBarEl = this.addStatusBarItem();
         this.statusBar = new StatusBar(statusBarEl);
-        this.setState(PluginState.status);
+        this.setState(PluginState.idle);
 
         const adapter: any = this.app.vault.adapter;
         const git = simpleGit(adapter.basePath);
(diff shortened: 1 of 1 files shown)


denolehov:
Looks clean! Thank you very much for your contribution! ❤️

kovetskiy:
@denolehov hey, you forgot to release the feature :smile: 
the latest version 1.4.0 doesn't include this auto push feature

denolehov:
@kovetskiy I didn't forget to release it. Unfortunately, I've encountered some visual bugs when I was testing things out and didn't get to fix them yet. :(

kovetskiy:
@denolehov I see. I am also diving into the source code of the plugin right now. How do you debug it? I mean, `console.log` doesn't work, is there any way to output something to console except like writing to a file?

denolehov:
@kovetskiy `console.log()` does work, to see the output, you need to open dev console via `ctrl + shift + I` shortcut.

Visual glitches I was referring to are related to race conditions when displaying messages. [Current implementation](https://github.com/denolehov/obsidian-git/blob/d0e0fbe4b7f3f3dde03b8d9496520038bbf1fba9/main.ts#L419-L428) works for simple cases, but it gets "confused" when there is a chain of messages to display. This needs to be refactored to use some sort of messages queue.

I.e. whenever we need to display a message, we add it to the list of messages to display, and then pop messages from there one by one and update status bar.

PR #47, opened 2021-02-17
Make git remote a configurable setting.

jsparkes:
I had several remotes defined in my global .gitconfig.  obsidian git was doing a "git remote" to get the remote branch name.  In my case, it got three names, separated by newlines.  When it tried to push to that name, it failed.

I have changes that make the git remote a configurable setting instead.  It defaults to "origin", but can be set to any legal remote name.  I currently have it set to "git@github.com:jsparkes/obsidian.git" which solves my problem.  (Using "origin" would work just as well, I was testing the absolute one.)

I think this is a flexible enough solution.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -52,14 +52,6 @@ export default class ObsidianGit extends Plugin {
         let branchInfo = await git.branch();
         this.settings.currentBranch = branchInfo.current;
 
-        let remote = await git.remote([]);
-        if (typeof remote === "string") {
-            this.settings.remote = remote.trim();
-        } else {
-            this.displayMessage("Failed to detect remote.", 0);
-            return;
-        }
-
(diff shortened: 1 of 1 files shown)


qubist:
For anyone else looking for a hot-fix for this problem of Obsidian Git trying to push to the wrong remote: in command-line, use `git push -u <name of remote>` to set the upstream branch. This way, `git push` by default will push to that remote. After a restart of Obsidian, Obsidian Git should push to that same remote.

Vinzent03:
An upstream branch is needed now for other checks. So I don't think the remote must be configurable in the settings. If you think so though, let me know.

Vinzent03:
I implemented it now differently myself in version 1.11.0

Issue #70, opened 2021-04-29
Option to hide the status bar entry?


OliverBalfour:
First things first, thanks for making this plugin, it's super helpful!

Just one thing, I'd like to be able to hide the git status bar entry. My status bar is quite cluttered and the git status bar changes state a lot which is distracting. Generally my network connection is good enough indication of the sync status so it doesn't add much.

I'd do this locally with CSS but there aren't any CSS selectors that differentiate the status bar entries from each other, only their innerHTML is different.

OliverBalfour:
Thanks!

Issue #68, opened 2021-04-20
Backup interval should outlive sessions


NomarCub:
I want to make sure to save daily, or a couple of times a week, but I don't have Obsidian open for that long, so the automatic vault backup doesn't kick in in time. To consistently save across sessions, the plugin could save the last backup timestamp into a file, and then auto backup on startup, if the interval has passed, or set a timer until the next scheduled backup time.
This could also help with more consistent saving if saving on exit (https://github.com/denolehov/obsidian-git/issues/13) is not possible.

Vinzent03:
Great idea! So you have set the time to something like 12 hours?

NomarCub:
I've set it to 10000 minutes and for a while forgot that it even existed.

PR #66, opened 2021-04-15
Mobile

Vinzent03:
# Checklist
- [x] commit
- [x] push
- [x] pull
- [x] status
- [x] error handling
- [x] validation check
- [x] Repository init
- [x] settings pane with configurations
- [x] disable status bar on mobile
- [x] port `simple-git` to new plugin structure
- [x] documentation
- [ ] community security check (I want to check the secure use of the personal access-token)

PR #71, opened 2021-05-02
feat: add env var OBSIDIAN_GIT for scripting

zanetumg:
I have a pre-push hook that ask for user confirmation. Sadly it makes obsidian-git push failed by timeout since I cannot make tty input.

A simple solution is to add an environment variable when spawn git process, check it in the hook and bypass the confirmation.

I also think configurable push command will be cool, but it's too much for this case.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -277,7 +277,7 @@ export default class ObsidianGit extends Plugin {
 
     async push(): Promise<void> {
         this.setState(PluginState.push);
-        await this.git.push(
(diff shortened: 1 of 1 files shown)


denolehov:
lgtm, @Vinzent03 agree?

Vinzent03:
So far I didn't work on other issues, because I didn't want to get merge conflicts with #66 , but I am already waiting a long time for the missing method from the Obsidian devs. I have no clue, when they will add that. I am okay with merging this. Will implement #70 and #68 hopefully this evening and will publish a release then.

PR #96, opened 2021-08-03
Status text formatting update

ronilaukkarinen:
There should be three dots in ongoing actions and no dots in last updated messages. Those two dots bug me surprisingly lot as an OCD person.


Vinzent03:
Hey, thanks for your contribution. To be honest, I didn't notice it myself. I don't know why `denelehov` choose two dots instead of three.

Issue #69, opened 2021-04-21
[Feature Request] Run on background process. / [Bug] UI freeze when run.


lkhphuc:
I set git backup to run automatically every hour.
Everytime it runs, Obsidian becomes laggy and unresponsive until it finish, which can takes up to 30s-60s. If I happen to be using Obsidian at this exact time 2-3 times in a morning, it can be very frustrating.
I wonder if this process can be run in the background? 

Maybe related if #68 is fixed.

MrVPlusOne:
I'm having the same issue here! One thing that feels strange to me is that if I manually push the changes from within the command line using `git`, it takes almost no time for the push to finish, whereas if I run the push command of the plugin from inside Obsidian, it freezes the UI and can take more than 30 seconds. What extra work does the plugin do that makes it so much slower than `git`?

23tux:
Same issue here! I've configured obsidian-git to push changes every 5 minutes, so the lagging / freezing becomes REALLY annoying.

broven:
Same issue

dsteinbock:
Same issue. When Obsidian Git is running, the Obsidian process spikes to 90% CPU for ~30 seconds.

When What OS are y'all running? I didn't notice this until after I upgraded from MacOS Catalina to Big Sur last week.

MrVPlusOne:
Actually, I think this issue was greatly improved after I upgraded my Obsidian and the plugin to the newest version. Maybe you can try to upgrade your Obsidian under the "Options -> About" tab.

akselw:
I have the same issue, and I experience it on both my Mac running Catalina and my Mac running Big Sur. I just noticed that Ben Hong had the same issue on his live stream a couple of days ago. You can see it, and he points it out at the 32:30 mark here: https://www.twitch.tv/videos/1044660956?t=0h32m30s

lkhphuc:
I updated the issue's name to reflect two side of this issue.

gabros20:
I have the same issue on Big Sur. Freezes the UI until the commit runs for 10-15 seconds. Recently I've updated Mac OS High Sierra to Big Sure and thats when I noticed the issue straight after.

dwSun:
Same issue,
Big Sur 11.4
Obsidian v0.12.5
Obsidian Git 1.9.2

I have 1625 files, most are pdfs, all those files take 1.7GB on my disk.

UI freezes for 10~15 seconds when starting up or pushing to github.

[...]

PR #108, opened 2021-09-05
Fix status bar "last update X ago"

denolehov:
@Vinzent03 I've noticed that we've stopped displaying time since the last update in the status bar, this PR fixes that.

Also, there were some typescript compilation errors, I've corrected those as well :)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -189,6 +189,7 @@ export default class ObsidianGit extends Plugin {
             }
         }
 
+        this.lastUpdate = Date.now();
         this.setState(PluginState.idle);
     }
 
@@ -248,6 +249,7 @@ export default class ObsidianGit extends Plugin {
                     return;
                 } else {
                     const pushedFiles = await this.gitManager.push();
+                    this.lastUpdate = Date.now();
                     this.displayMessage(`Pushed ${pushedFiles} files to remote`);
                 }
             } else {


denolehov:
I am wondering if those typescript errors are related to #69 since `simple-git` was receiving a callback for `options`.

src/main.ts
```diff
@@ -92,8 +92,7 @@ export default class ObsidianGit extends Plugin {
                 window.setInterval(() => this.statusBar.display(), 1000)
             );
         }
-        this.app.workspace.onLayoutReady(() => this.init());
```
Vinzent03, line 95: I guess you are on an old version of the api. Try running `npm update`. Your fix is actually deprecated.

src/simpleGit.ts
```diff
@@ -89,7 +89,7 @@ export class SimpleGit extends GitManager {
         if (this.plugin.settings.updateSubmodules === true) {
             return true;
         }
-        const status = await this.git.status((err: any) => this.onError(err));
+        const status = await this.git.status([], (err: any) => this.onError(err));
```
Vinzent03, line 92: This should not be needed, because there is a method definition with just the error callback. `simple-git.d.ts`:
```ts
/**
    * Show the working tree status.
    */
   status(options?: types.TaskOptions, callback?: types.SimpleGitTaskCallback<resp.StatusResult>): Response<resp.StatusResult>;

   status(callback?: types.SimpleGitTaskCallback<resp.StatusResult>): Response<resp.StatusResult>;
```

src/simpleGit.ts
```diff
@@ -119,15 +119,15 @@ export class SimpleGit extends GitManager {
...
+        await this.git.checkout(branch, [], (err: any) => this.onError(err));
     };
 
     async init(): Promise<void> {
         await this.git.init(false);
     };
 
     async setConfig(path: string, value: any): Promise<void> {
-        await this.git.addConfig(path, value, (err: any) => this.onError(err));
+        await this.git.addConfig(path, value, false, (err: any) => this.onError(err));
```
Vinzent03, line 130: Same as with `status`

Vinzent03 (changes requested):
Oh you are right with updating last update. I must somehow miss that. Thanks for the fix. But I don't agree with the other changes.

src/main.ts
denolehov: oh, you're right! I've ran `npm install` and assumed that I am up-to-date :facepalm:

src/simpleGit.ts
denolehov: running `npm update` made the errors go away :+1: I have dropped this commit

Vinzent03:
Looks good :joy: Will create new release right after merge. Thanks for noticing.

PR #127, opened 2021-10-28
Add Git View, migrate to ESBuild

phibr0:
- [x] ESBuild Config
- [x] Sidebar View
     - [x] Hover Preview
     - [x] Path in aria-label
     - [ ] Logic for all the Buttons

...


miluoshi:
> I just did it because npm's `package-lock.json` is also ignored by default (in the sample plugin). Is there any difference?

I'm not sure why was `package-lock.json` ignored, but the point is there should be just one of those files depending on whether project uses Yarn or NPM and it should be checked in.

PR #156, opened 2021-12-21
fix: Message format grammar plurality

ethyi:
Added conditions to change message depending on if there is one or multiple files.


Vinzent03:
LGTM! Thanks for your contribution.

PR #159, opened 2021-12-25
speedup obsidian start

luckydevil13:
just wrap onload within requestIdleCallback

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -37,110 +37,113 @@ export default class ObsidianGit extends Plugin {
         await this.loadSettings();
         addIcons();
 
-        this.registerView(VIEW_CONFIG.type, (leaf) => {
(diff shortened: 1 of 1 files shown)


Vinzent03:
I don't know what you want to speed up there. All those tasks you wrapped aren't heavy and take much time. The heavy tasks like checking for git installation are already wrapped up in `onLayoutReady()`, which is called when Obsidian is finished with starting and the editor is shown.

phibr0:
By adding `require("electron").remote.getCurrentWindow().close()` to the end of the init function and then measuring the time Obsidian needs to start up, I get the following times:

**Default:** 2.09 seconds
**This PR:** 2.17 seconds
(Median values of 5 separate runs each)

Vinzent03:
Will close this until you can argue your changes.

PR #165, opened 2022-01-17
Add new 'Sync Method' setting in place of `mergeOnPull` and new experimental obsidian-sync compatible sync method.

coddingtonbear:
I use both obsidian-sync and obsidian-git so I can have both live updates & mobile support (via obsidian-sync) while also having change-by-change history (via obsidian-git).  While far from home over the holidays with my work laptop, I encountered more than a few merge conflicts due to a race condition in which my laptop at home dutifully received changes via obsidian-sync, committed them, and then while it attempted to push those changes, it discovered that my work laptop had already committed changes and was thus in a conflict.

What this does is add a third method for handling sync (in addition to the existing `rebase` and `merge` methods) that just lazily updates the current `HEAD` to match the sha of any fetched remote data *without updating the local filesystem*.  That might seem surprising at first glance, but while using obsidian-sync, this works perfectly well due to obsidian-sync itself updating your local filesystem automatically as changes are discovered -- almost universally long before obsidian-git has even had a chance to discover that those changes exist on the remote.  Additionally, if we keep obsidian-git _out_ of that process, it can perform its normal commit/push processes for keeping an accurate historical record of changes to your notes without the possibility of getting trapped in a conflict.

This also refactors _how_ we indicate to obsidian-git which sync method to use.  Before this change, there was a setting named `mergeOnPull` that is either `true` (indicating that we should merge fetched changes) or `false` (indicating that we should rebase atop fetched changes).  After this change, there is now a setting named `syncMethod` that can be set to one of three values: `merge`, `rebase`, or `reset` (this is the new obsidian-sync compatible option).  Settings will be automatically migrated forward at plugin load-time.

I've been using this successfully for a week or so now, but it's important for me to say that although I've tested this pretty thoroughly with the new `reset` method and spot-checked `merge` and `rebase`, it would be best if somebody who uses the more-normal sync methods gave this a shot, too, before merging.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -37,6 +37,7 @@ export default class ObsidianGit extends Plugin {
     async onload() {
         console.log('loading ' + this.manifest.name + " plugin");
         await this.loadSettings();
+        this.migrateSettings();
 
         addIcons();
 
@@ -159,7 +160,6 @@ export default class ObsidianGit extends Plugin {
             }
         });
 
-        this.migrateSettings()
 
         if (this.settings.showStatusBar) {
             // init statusBar
@@ -174,10 +174,10 @@ export default class ObsidianGit extends Plugin {
     }
 
     migrateSettings() {
-        if(this.settings.mergeOnPull) {
-            this.settings.syncMethod = 'merge'
-            this.settings.mergeOnPull = undefined
-            this.saveSettings()
+        if (this.settings.mergeOnPull != undefined) {
+            this.settings.syncMethod = this.settings.mergeOnPull ? 'merge' : 'rebase';
+            this.settings.mergeOnPull = undefined;
+            this.saveSettings();
         }
     }
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -158,6 +158,9 @@ export default class ObsidianGit extends Plugin {
                 new ChangedFilesModal(this, status.changed).open();
             }
         });
+
+        this.migrateSettings()
+
         if (this.settings.showStatusBar) {
             // init statusBar
             let statusBarEl = this.addStatusBarItem();
@@ -168,6 +171,14 @@ export default class ObsidianGit extends Plugin {
         }
         this.app.workspace.onLayoutReady(() => this.init());
 
+    }
+
+    migrateSettings() {
+        if(this.settings.mergeOnPull) {
+            this.settings.syncMethod = 'merge'
+            this.settings.mergeOnPull = undefined
+            this.saveSettings()
+        }
     }
 
     async onunload() {
@@ -386,11 +397,7 @@ export default class ObsidianGit extends Plugin {
         const pulledFilesLength = await this.gitManager.pull();
 
         if (pulledFilesLength > 0) {
-            if (this.settings.mergeOnPull) {
-                this.displayMessage(`Pulled ${pulledFilesLength} ${pulledFilesLength > 1 ? 'files' : 'file'} from remote`);
-            } else {
-                this.displayMessage("Rebased on pull");
-            }
+            this.displayMessage(`Pulled ${pulledFilesLength} ${pulledFilesLength > 1 ? 'files' : 'file'} from remote`);
         }
         return pulledFilesLength != 0;
     }
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -91,7 +91,7 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
                 const options: Record<SyncMethod, string> = {
                     'merge': 'Merge',
                     'rebase': 'Rebase',
-                    'reset': 'None (for use with Obsidian Sync)',
+                    'reset': 'Other sync service (Only updates the HEAD without touching the working directory)',
                 };
                 dropdown.addOptions(options);
                 dropdown.setValue(plugin.settings.syncMethod);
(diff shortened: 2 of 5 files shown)


src/simpleGit.ts
```diff
@@ -146,33 +147,43 @@ export class SimpleGit extends GitManager {
         this.plugin.setState(PluginState.pull);
         if (this.plugin.settings.updateSubmodules)
             await this.git.subModule(["update", "--remote", "--merge", "--recursive"], (err: any) => this.onError(err));
-        let lastRemoteCommitBefore: string;
-        if (!this.plugin.settings.mergeOnPull) {
-            lastRemoteCommitBefore = await this.getNewestRemoteCommit();
 
-        }
-        const pullResult = await this.git.pull([this.plugin.settings.mergeOnPull ? '--no-rebase' : '--rebase'],
```
coddingtonbear, line 154: Previously, we were using `git pull` to fetch and merge changes.  `git pull` is a shorthand for `git fetch` followed by either `git rebase FETCH_HEAD` or `git merge FETCH_HEAD` (https://git-scm.com/docs/git-pull/2.1.4#_description).

In the new changes on the right, you'll see that we're intentionally `git fetch`-ing separately from `git merge` or `git rebase`, but otherwise these are largely functionally the same.

src/main.ts
```diff
@@ -174,10 +174,10 @@ export default class ObsidianGit extends Plugin {
     }
 
     migrateSettings() {
-        if(this.settings.mergeOnPull) {
-            this.settings.syncMethod = 'merge'
-            this.settings.mergeOnPull = undefined
-            this.saveSettings()
+        if (this.settings.mergeOnPull != undefined) {
```
Vinzent03, line 177: We have to set `syncMethod` on `rebase` too, don't we?

src/settings.ts
```diff
@@ -87,22 +87,19 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
...
                 const options: Record<SyncMethod, string> = {
                     'merge': 'Merge',
                     'rebase': 'Rebase',
                     'reset': 'None (for use with Obsidian Sync)',
-                }
-                dropdown.addOptions(options)
-                if (plugin.settings.syncMethod) {
-                    dropdown.setValue(plugin.settings.syncMethod)
-                } else {
-                    dropdown.setValue('merge')
```
Vinzent03: Because of the default value of `syncMethod` it should always be non undefined.

src/simpleGit.ts
```diff
@@ -148,40 +147,42 @@ export class SimpleGit extends GitManager {
...
 
         if (localCommit !== upstreamCommit) {
             if (this.plugin.settings.syncMethod === 'merge' || this.plugin.settings.syncMethod === 'rebase') {
                 try {
-                    switch(this.plugin.settings.syncMethod) {
+                    switch (this.plugin.settings.syncMethod) {
                         case 'merge':
-                            this.git.merge([branchInfo.tracking])
-                            break
+                            await this.git.merge([branchInfo.tracking]);
```
Vinzent03, line 161: await is needed to catch the error.

src/types.ts
```diff
@@ -16,11 +16,13 @@ export interface ObsidianGitSettings {
     autoBackupAfterFileChange: boolean;
     treeStructure: boolean;
 
-    /* Obsolete settings */
-    mergeOnPull?: boolean;  // Migrated to `syncMethod = 'merge'`
+    /**
+     * @deprecated Migrated to `syncMethod = 'merge'`
```
Vinzent03, line 20: I think using `@deprecated` is even better to support highlighting its usage in the editor.

Vinzent03 (commented):
Sorry for my late review. I like the idea to solve this race condition. I noticed that files edited on another device are always staged. Did you experience this too? What do you think about my changes?

src/main.ts
coddingtonbear: I don't _believe_ so given that `rebase` is the default, but I'm :+1: on making this setting explicit so this looks reasonable to me.

coddingtonbear: That all being said -- if we're going the explicit route, we probably need to make sure this ends up being set even if `this.settings.mergeOnPull` isn't set, no?  Otherwise, I'm pretty sure this particular setting will still possibly be `undefined` for users that generated their settings before this PR.

src/simpleGit.ts
coddingtonbear: Oo; good catch!

src/types.ts
coddingtonbear: I had no idea that even existed, but yeah, that's clearly better.

src/main.ts
Vinzent03: > I don't believe so given that rebase is the default,

No, because the default of `mergeOnPull` is/was `true`

> ends up being set even if this.settings.mergeOnPull isn't set, no?

https://github.com/denolehov/obsidian-git/blob/master/src/main.ts#L182 sets values from the default config for undefined keys in the loaded setting. So this shouldn't be a problem.

coddingtonbear: Oh, neat; I had totally missed that!

Vinzent03:
LGTM now. Thanks for the (I think) great idea of solving race conditions with other sync services.

coddingtonbear:
Awesome -- thanks for merging.  Just btw, I'd been meaning to look into your comment "I noticed that files edited on another device are always staged", but haven't yet found the time and intend to have a closer look sometime over the next few days.

coddingtonbear:
Oh, nevermind, I see that you fixed that in f60f95c; thanks!

Vinzent03:
Yeah I'm still unsure, but I think just unstaging all is fine.

PR #188, opened 2022-02-24
feat: add new background auto backup mode

miluoshi:
This change introduces new auto backup mode discussed in https://github.com/denolehov/obsidian-git/issues/69#issuecomment-1028173943 that is triggered after X minutes of Obsidian window in background. If Obsidian window is focused while the timer is running, it is reset and new timer starts running only after Obisidian window is put to background again.

The new dropdown setting `autoSaveIntervalMode` also replaces existing toggle setting `autoBackupAfterFileChange` which is now deprecated.

Values of old toggle setting are migrated like this:
- `autoBackupAfterFileChange = false` -> `autoSaveIntervalMode = 'default'`
- `autoBackupAfterFileChange = true` -> `autoSaveIntervalMode = 'after-change'`

New mode has value `'after-inactive'`.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -4,6 +4,7 @@ export const DEFAULT_SETTINGS: ObsidianGitSettings = {
     commitMessage: "vault backup: {{date}}",
     commitDateFormat: "YYYY-MM-DD HH:mm:ss",
     autoSaveInterval: 0,
+    autoSaveIntervalMode: 'default',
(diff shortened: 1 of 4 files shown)


Vinzent03:
I see that I never responded here. Sorry. Someone stated some problems with this pr so I didn't review it. Since #69 should be resolved now, I'll close this pr.

Issue #169, opened 2022-01-22
[Idea / Question] Commit part of a vault, not the whole vault


Mynotoar:
I want to sync a folder within my vault across multiple devices, for which Obsidian Git sounds like the neatest solution. But as far as I can tell, you can only sync the whole vault. Have I missed something - is there a way to sync part of a vault? In the same way that in any other working directory I could just selectively add files, not the whole thing.

If it's not possible, then this is a feature request / idea. If it is possible I'd appreciate a pointer. Cheers.

miluoshi:
You are free to ignore any folders or files using `.gitignore` file - that's one option. Other one to achieve it is to open a repository in a subfolder of your vault that you want to sync through git.

Issue #11, opened 2020-11-05
Allow specifying custom repository root


denolehov:
Expose a setting for specifying a custom repository root in order to cover [use-case](https://github.com/denolehov/obsidian-git/issues/7#issuecomment-722032481) where git repo is not in the root of the vault, but in one of the subfolders that users wish to backup.

dummifiedme:
Related to this

My use case is my github repo is say Folder A, and I want a subfolder B to be my Vault. Will it work if I work on folder B from Obsidian? or the I have to compulsarily make Folder A as my Vault.

denolehov:
As of right now, you'd have to make Folder A as your vault. Plugin looks for git repo in the root of the vault.

PR #209, opened 2022-03-20
feat: custom git repository root

mbseid:
## Context

For my use case, I wanted to have a public blog live within my obsidian vault. This blog is hosted and built on github pages and is a normal git repository. To keep things separated, the git repository be in a subfolder not at the root of the vault. This is unsupported currently and referenced in a few issues:

* https://github.com/denolehov/obsidian-git/issues/11 
* https://github.com/denolehov/obsidian-git/issues/169

## Solution

A setting was added called bath path that allows users to set a custom git root. This is then used by `simple-git` when configuring the SimpleGit RepoManger. 

Screen Shot 2022-03-20 at 11 14 04 AM

## Limitations

A user is only allowed on repo in the entire vault. I evaluated having multiple git repo support but it turned out to be a larger refactor than I had time for. For my use case this is totally fine. 

Thanks for the great plugin!

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -306,7 +306,8 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
             });
 
         new Setting(containerEl)
-            .setName("Base Path")
+            .setName("Git Repository Path")
+            .setDesc("Sets the relative path to the git repository.")
             .addText((cb) => {
                 cb.setValue(plugin.settings.basePath);
                 cb.onChange((value) => {
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -304,6 +304,17 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
                     plugin.gitManager.updateGitPath(value || "git");
                 });
             });
+
+        new Setting(containerEl)
+            .setName("Base Path")
+            .addText((cb) => {
+                cb.setValue(plugin.settings.basePath);
+                cb.onChange((value) => {
+                    plugin.settings.basePath = value;
+                    plugin.saveSettings();
+                    plugin.gitManager.updateBasePath(value || "");
+                });
+            });
         const info = containerEl.createDiv();
         info.setAttr("align", "center");
         info.setText("Debugging and logging:\nYou can always see the logs of this and every other plugin by opening the console with");
--- a/src/simpleGit.ts
+++ b/src/simpleGit.ts
@@ -20,9 +20,14 @@ export class SimpleGit extends GitManager {
         if (this.isGitInstalled()) {
             const adapter = this.app.vault.adapter as FileSystemAdapter;
             const path = adapter.getBasePath();
-            const extraPath = this.plugin.settings.basePath || ""
+            let extraPath = ""
+            // Because the basePath setting is a relative path, a leading `/` must
+            // be appended before concatenating with the path.
+            if (this.plugin.settings.basePath){
+              extraPath = "/" + this.plugin.settings.basePath;
+            }
             this.git = simpleGit({
-                baseDir: path + "/" + extraPath,
+                baseDir: path + extraPath,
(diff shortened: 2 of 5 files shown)


src/simpleGit.ts
```diff
@@ -20,8 +20,10 @@ export class SimpleGit extends GitManager {
         if (this.isGitInstalled()) {
             const adapter = this.app.vault.adapter as FileSystemAdapter;
             const path = adapter.getBasePath();
+            const extraPath = this.plugin.settings.basePath || ""
+            console.log(path + extraPath)
```
Vinzent03, line 24: Please remove that `console.log()`

src/settings.ts
```diff
@@ -304,6 +304,17 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
...
             });
+
+        new Setting(containerEl)
+            .setName("Base Path")
+            .addText((cb) => {
+                cb.setValue(plugin.settings.basePath);
+                cb.onChange((value) => {
+                    plugin.settings.basePath = value;
+                    plugin.saveSettings();
+                    plugin.gitManager.updateBasePath(value || "");
```
Vinzent03, line 315: I would first apply the new path and then save the setting. So when updating the base path fails, the setting doesn't get saved.

src/settings.ts
```diff
@@ -304,6 +304,17 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
                     plugin.gitManager.updateGitPath(value || "git");
                 });
             });
+
+        new Setting(containerEl)
+            .setName("Base Path")
+            .addText((cb) => {
```
Vinzent03, line 310: A description of what `Base Path` is, is needed. Like that, it needs to be relative to the vault. In addition, I would rename it to something like `git repository path`.

src/simpleGit.ts
```diff
@@ -20,8 +20,10 @@ export class SimpleGit extends GitManager {
...
             this.git = simpleGit({
-                baseDir: path,
+                baseDir: path + extraPath,
```
Vinzent03, line 26: `extraPath` has to start with an `/`, doesn't it? Did you specify `extraPath` with a leading `/` in your testing? That's weird, because it's a relative path.

Vinzent03 (changes requested):
Thanks for your contribution. I left you some comments.

src/settings.ts
mbseid: I think the settings need to be saved first. The `updateBasePath` function doesn't actually use the arguments provided. This follows the same path as the Git Path.

https://github.com/mbseid/obsidian-git/blob/basePath/src/simpleGit.ts#L314-L320

mbseid:
Great feedback. I updated the setting to add more clarity and to use relative paths. I did use an absolute path during development. I agree that it makes more sense for it to be a relative path.

src/simpleGit.ts
mbseid: Great feedback. I updated the plugin to use relative paths.

src/settings.ts
Vinzent03: Ah good catch. Then keep your changes.

src/simpleGit.ts
Vinzent03: I added a commit myself using platform specific separator, because Windows uses `\`.

Vinzent03:
When testing locally, I noticed that you forgot to add `basePath` to the settings interface. Did you just forget to commit those changes, or did you ignore the analyzer? Nevertheless, I fixed that.

Vinzent03 (approved):

mbseid:
Awesome, thanks for your help. I didn't see anything in the analyzer, I may have missed it though. Also nice catch on the separator. Haven't written anything cross platform before and didn't think about the separator.  Let me know if you need anything else to merge, I'm not able to because I don't have write access.

PR #237, opened 2022-05-18
add disable commit hooks setting

nickyhuyskens:


Vinzent03:
Could you explain why you need this option? E.g. why don't disable hooks in this repo completely?

nickyhuyskens:
On my work computer, we have global git commit hooks that work on all repos and disallow commiting to master.
My obsidian repo is self-hosted at home and my main branch is master so this plugin would not commit and push.

Vinzent03:
I think this use case it a bit too specific for an extra setting option. When you are fine with disabling hooks completely for that git repo, you could try [this answer](https://stackoverflow.com/a/67527118). Of course omit `--global`.

Issue #260, opened 2022-07-12
Enhancement request: Source Control View lists recently pulled files


peterkaminski:
## Current Behavior

After doing a pull, Obsidian Git currently displays a pop-up message like this:

> Pulled new changes. 10 files updated

It would be nice to know _which_ files were updated.

## Idea: pulled files listed in Source Control View

Idea: Use the bottom half of the Source Control View pane to display the names of the files that were updated in the most recent pull.

Mockup:

![Source Control View lists recently pulled files mockup 2022-07-12](https://user-images.githubusercontent.com/169442/178581714-ff4b757f-2353-4eb5-8f3b-f60e23013ed5.jpg)

## Technical considerations: `git pull`

Part of the `git pull` output looks like this, and may be able to be parsed to get needed information:

```
Obsidian Tips and                                          | 165 +++++++++++++++++--------------------------
Decentralizing, Centralizing, and Decentralizing Again     |  46 ------------
Office Hours                                               | 198 ----------------------------------------------------
Connections (British Documentary)                          |  17 +++++
 ```
 
 Internally, `git pull` is a fetch and merge, offhand I don't know which one would generate the information above, or if there is a porcelain version of it.

Vinzent03:
I definitely like the idea, but I don't when I get the time for it.

PR #269, opened 2022-08-06
Implement `Recently Pulled Files` section

vbeskrovnov:
This PR implements #260
Each time the user pulls at least one file, the list of pulled files will be shown in the new section. After app restart, this list will be empty as we are not storing any history.

TreeView:
image

DefaultView:
image


Vinzent03:
Hey thanks for your contribution! I've made some changes to hide the thing when there are no recently pulled files and to update the view when e.g. pulled via command.
In case you are fine with the changes, we can merge this!

PR #278, opened 2022-08-11
fixed grammatical error

dojje:
when there is 0 of something you use a plural noun. sometimes you commit 0 files, like when you resolve conflicts it says "committed 0 file". it was a bit annoying so i fixed it.


Vinzent03:
Did you encounter this issue, because it shouldn't get fired when there are no changes. Before commit, it checks for count of index and non index file changes.

dojje:
when you have commited files but you haven't pushed them it says committed 0 file.

Vinzent03:
This got now resolved in #393

PR #288, opened 2022-08-27
Line-wise author info (resolved #77)

GollyTicker:
This is a **work in progress PR!**

@Vinzent03 Could you take a look at the first two commits only?
For working with the plugin, I've added a dockerized environment to reproducibly setup the node packages. I also added a [test-vault](https://github.com/GollyTicker/obsidian-git-test-vault-online) as a submodule. 
If you think, they are useful, I can extract them into separate branches and have them merged into the plugin for reproducability and makiing it easier for others to contribute. Otherwise, I can simply remove them from the branch and keep them just to me for development purposes.
What do you think?

Regarding the actual implementation, I will ping in actual issue, when I have a relevant working state to disucss.

PR #320, opened 2022-09-22
Dockerized setup for development

GollyTicker:
As discussed with @Vinzent03, this PR contains the setup that can be used by new developers to quickly get into this plugin - as all they need to have is Docker. Simply running `scripts/restart-dev.sh` will install all dependencies and run the compiler in watch-mode.

I've extracted this code from #288.

This PR is a draft, since it was decided to not merge this. Hence, this is here primarily for future reference.

If you want, we can move this into an issue instead.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+node_modules
+.git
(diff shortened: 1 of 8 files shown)


Vinzent03:
I don't know docker that well and feel like nodejs is installed on almost every dev machine, so I don't really see the need as well.

Issue #73, opened 2021-05-09
Option take last commit time into account for backup 


NomarCub:
I'd like the plugin to optionally check the last commit time and update the last auto-backup time to that.
For example let's say the last auto-backup was 30 minutes ago, the interval is an hour and I make a manual commit now. With this option auto-backup won't trigger 30 minutes from now, but an hour from now, an hour from the last commit.

Vinzent03:
I see your point, but honestly that's not really what I understand under an auto backup.
What's your opinion @denolehov ?

denolehov:
The plugin's auto-backup internal timer is reset once you manually commit from within Obsidian. Tracking commits made from outside of Obsidian would be overkill.

Vinzent03:
@denolehov Are you sure? Did I remove that with my changes? I can only find the update of `lastUpdate`, but that's only for the status bar.

denolehov:
Personally, I am fine with either having a backup timer reset on manual commit from within Obsidian UI or not having it reset at all and have it free-running.

But in any case, having a reset timer rely on git history (i.e. periodically fetching git log and seeing what was the latest timestamp) is definitely an overkill.

But maybe it's just me, and if more people request this change, we could perhaps implement a "Backup Mode" where people would be able to choose whether they want to have the timer free-running or be based on the latest commit, or be reset on manual commits from Obsidian.

NomarCub:
I imagined the plugin wouldn't need to check the git log periodically, only on startups and before scheduled backups, like with `loadLastAuto()` in `init` now.
Are you open to me implementing this in the future?

NomarCub:
For anyone still interested, I found a workaround with git hooks. It works well for me on Windows.
I made a post commit hook (in `.git/hooks/post-commit`), that overwrites the plugin's data file with the current commit time, like so:
```
#!/bin/bash
stamp=$(date +"%Y-%m-%dT%T%z")
echo $stamp
echo -e "$stamp\n$stamp" > .obsidian-git-data
```

This makes the shell print [the current date](https://www.cyberciti.biz/faq/unix-linux-getting-current-date-in-bash-ksh-shell-script/) in JavaScript's [Date.parse()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#date_time_string_format) format used by the plugin.

NomarCub:
[...]

PR #335, opened 2022-10-11
git log on GitManager interface

NomarCub:
Is this what you meant [here](https://github.com/denolehov/obsidian-git/issues/73#issuecomment-1273049869) @Vinzent03?
I left a todo where the mobile implementation should be. You might want to refactor the placement and the types.
Alternatively, for my specific use case a `lastCommitDate()` function on GitManager would be ideal, but I don't know if you'd want that on the interface.

### DIFF
--- a/src/gitManager.ts
+++ b/src/gitManager.ts
@@ -211,4 +211,8 @@ export abstract class GitManager {
         }
         return template;
     }
+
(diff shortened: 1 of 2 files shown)


NomarCub:
@Vinzent03 how long should I expect this to take? Is there a lot of modifications needed for it?

Vinzent03:
Thanks for that start, but I implemented it now completely myself.

PR #355, opened 2022-10-28
Update Description for Auto Backup-Toggle

keeev:
For some reason, I wasn't able to run this locally so I had to do it on GitHub itself, I hope my changes in the settings make sense. 

I was confused by the setting for auto backups for file changes, because of that I added a name and changed the description so it should be clear. 

Does something happen when the toggle is `off` if not, then the "If turned off, ..." was redundant anyway?! 

If you need anything else, let me know. Great plugin by the way. 🙌🏼

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -70,7 +70,8 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
                 );
 
             new Setting(containerEl)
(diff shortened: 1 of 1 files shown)


Vinzent03:
Thanks for the contribution! I added `if turned off`, because it's not like I turn something on, rather switch between two different modes. But adding it as description is definitely better than writing everything in the title.

Issue #292, opened 2022-09-02
[bug] `DisableNotification` setting does not work on obsidian git mobile (iOS).


andylizf:
Maybe because there is no check for the `DisableNotification` setting before raise a `Notice` in file [isomorphicGit.ts](https://github.com/denolehov/obsidian-git/blob/master/src/isomorphicGit.ts)

Vinzent03:
Some commands can take a while. So I want the user
always to know that the command is still running. I may delay the notice so that they come only up if the task takes like > 5 seconds. Which notices are disturbing you in particular?

mwxgaf:
It shows **all** notifications even if you disable notifications (In Android too)

Vinzent03:
@mwxgaf Notifications like
- Everything is up to date

mwxgaf:
yes yes
this is right
I had not paid attention
this is not about all notifications

> You should see notifications like Getting Status or progress notifications

do you have any plan to fix this?

green-codes:
The spec is that only *errors* are shown, *all* other notifications should be silenced. 

I depend on Obsidian Git for syncing all my vaults and have all auto backup options on, and it’s quite annoying to be bombarded by status updates every time I startup the iOS app. Hope this is fixed soon.

mwxgaf:
For a temp solution I commented all `Notice` lines in the JS file

Vinzent03:
Should be fixed now. If the status command takes longer than 20s I'm still showing a notification to let the user know a task is running long.

bd-g:
I am getting an error with the new pull request

```
TypeError: undefined is not an object (evaluating progressNotice.noticeEl')
```

bd-g:
@Vinzent03 just implementation details - for example, in file `obsidian-git/src/isomorphicGit.ts`, a sample change is below

```
// This used to ALWAYS be a Notice object - now it can be undefined
const progressNotice = this.showNotice("Initializing pull");
...
...
// progressNotice is used here, but if it is undefined it will fail
(progressNotice as any).noticeEl.innerText = this.getProgressText("Checkout", progress);
```

I will see if there is a fix in the next day or two. If not, I may try to find time to make a PR with my own fix.

echo-saurav:
Am i suppose to see "finish pull" every time?? . i get this notification on ios

Vinzent03:
@echo-saurav Should be fixed now!

PR #363, opened 2022-11-03
Fix bug in hiding mobile notifications

bd-g:
Further fix for #292 

There was a fix made in commit `7d6252795ca62f4176fee90d674041659a0a1d9f`, however, this introduced a bug where "progressNotice" could be undefined, and the progress function was attempting to set a property on this undefined object. This would throw an error, and many git operations could not be performed.

I simply wrapped the progressNotice in an if statement to check if it was defined - without tests I am not sure if there are side effects, but this might work.

### DIFF
--- a/src/isomorphicGit.ts
+++ b/src/isomorphicGit.ts
@@ -303,7 +303,9 @@ export class IsomorphicGit extends GitManager {
                 ...this.getRepo(),
                 ref: branchInfo.current,
                 onProgress: (progress) => {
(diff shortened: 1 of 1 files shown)


bd-g:
Note for @Vinzent03 , v2.9.2 is currently broken for mobile, so need either a rollback or to accept this PR

Especially since Obsidian doesn't have a way to specify which version of an extension you want to target...

Thanks for taking a look at this!

Vinzent03:
Thank you very much for that quick fix! That's the danger of `as any`, isn't it?

PR #409, opened 2022-12-27
Adds option to recursively checkout submodules

DenEwout:
This commit allows users to automatically checkout the same branch in submodules (if present) as the root repository. This does not create any branches.

Additional refactor for retrieval of submodule paths to allow code reuse.

Additional question: we've noticed that the "tabs" indent_style setting in the .editorconfig does not match the indentation used in the code (being spaces). Which one of the two should it be?

(co-authored by @yentlprojects)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -14,7 +14,6 @@ export interface ObsidianGitSettings {
     showStatusBar: boolean;
     updateSubmodules: boolean;
 	submoduleRecurseCheckout: boolean;
-	matchRootBranchName: boolean;
 	/**
     * @deprecated Using `localstorage` instead
     */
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -13,7 +13,9 @@ export interface ObsidianGitSettings {
     listChangedFilesInMessageBody: boolean;
     showStatusBar: boolean;
     updateSubmodules: boolean;
-    /**
+	submoduleRecurseCheckout: boolean;
+	matchRootBranchName: boolean;
```
Vinzent03, line 17: I think this setting is a left over?

DenEwout: Yup! Removed it, cheers.

Vinzent03 (changes requested):

Vinzent03:
I've fixed the `.editorconfig` file now. I'm using everywhere `space` now.

Vinzent03:
LGTM!

PR #419, opened 2023-01-07
Adding a license file.

jkamenik:
I have no opinion of what license you should be using, but I do think that each project (especially useful ones) should have a license.  I choose MIT as the default for my own stuff, but if you prefer a different license I would be happy to change it.

I updated the Copyright line to include your name (taken from your profile) and the year of the first commit (taken from `git log`).

Let me know if you want anything changed.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Denis Olehov
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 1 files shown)


Vinzent03:
Hmm, I'm maintaining this plugin for the last two years now, but I don't know if I'm theoretically allowed to decide that for him. So I guess I have to ask @denolehov. Is MIT okay for you?

denolehov:
MIT is okay 👍

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Denis Olehov
```
denolehov, line 3: @Vinzent03 we must add your name here as well, and it should be before mine :D

denolehov: Given that your handle is technically PII, it is possible to use it (according to my quick google search :D)

jkamenik:
I have updated to include both of you in the order requested.  Thank you.

Vinzent03:
LGTM! thanks

denolehov:
Thank you @jkamenik! 🙌

PR #443, opened 2023-02-23
Explain `submodule` feature needs to be enable explicitly

### LINKED ISSUE #442: [Bug]: Git submodule does not work when adding new files to submodule
### Describe the bug

Hi thanks for the great plugin! However submodules does not seem to work. More specifically:

* `~/a` is the obsidian vault, as well as git repo (with `~/a/.git` storing those git things). the remote of this git repo is (say) `github.com/a`
* `~/a/images` stores the large attachments, and is a `git submodule`. the remote of this git submodule repo is (say) `github.com/images`

The problem is, when I trigger "obsidian backup", it correctly commits and pushes for the `a` (i.e. main repo), but it just *completely ignores* the `a/images` submodule. In other words, the changed files in the submodule is never commited or pushed. This makes the plugin useless :(

More specifically, suppose we add `1.png` into `~/a/images/` folder. Then when triggering obsidian backup, I expect that the submodule gets a new commit with this new file, and the main git repo also gets a new commit with the changed submodule. But indeed nothing happens.

### Relevant errors (if available)

_No response_

### Steps to reproduce

see above

### Expected Behavior

_No response_

### Addition context

_No response_

### Operating system

macOS

### Installation Method

Other

### Plugin version

latest

fzyzcjy:
Close #442

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ For mobile users see [Mobile](#mobile) section below.
 - Automatic vault backup every X minutes
 - Pull changes from remote repository on Obsidian startup
 - Assign hotkeys for pulling/pushing changes to a remote repository
-- Manage different repositories via Git submodules
+- Manage different repositories via Git submodules (after enabling this feature in settings)
 - Sidebar view. The Source Control View allows you to stage and commit individual files. It can be opened with the `Open Source Control View` command. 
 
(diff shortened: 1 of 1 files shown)


Vinzent03:
Thanks for the clarification!

fzyzcjy:
You are welcome!

PR #445, opened 2023-02-25
Add error handlings when retrieving branch names

tamiroh:
Added error handlings when a remote branch is not configured.

Although I could not imagine a case where a remote branch name could be retrieved but not the current branch name, it could occur based on the types, so I also added.


Vinzent03:
LGTM, Thanks for your contribution!

Issue #340, opened 2022-10-16
[Bug]: Merge Conflicts not supported on mobile?


tjcampanella:
### Describe the bug

When attempting to pull on Obsidian mobile I get an error saying "MergeNotSupportedError: Merge with conflicts is not supported yet". I am unsure if this is a bug or it is just not supported yet like the message says. I saw an issue saying that merges should work on the Obsidian Git Mobile issue. If it is not supported then please let me know how I should handle this. Thanks.

### Steps to reproduce

I have automatic pull enabled so when I startup Obsidian it attempts to pull down the changes however, it fails with a MergeNotSupportedError. I am unsure exactly what I did to cause this I assume I made some changes on my phone without syncing with my computer so things must have gotten out of sync. 

### Operating system

iOS

### Plugin version

2.6.0

glassoff:
I have the same problem: I've got conflict on mobile and now can't make pull and as I right understand there are not in mobile ways to resolve conflicts :( 
At least it would be enough to have ability to open conflict file for manual conflict resolving.

bmcilw1:
I'm seeing this error as well and I'm not sure what to do

Vinzent03:
Sorry, but I can't reproduce. Are you sure you have the same root history? So the histories are not completely divergent?

bmcilw1:
Ah, I see that I had the merge strategy set to rebase for my desktop and merge on my phone. Git has problems when trying to mix strategies. Trying both to "merge" to see if that's better

J-Emmerich:
I was having the same issue with a fresh git install that I did in a new laptop. 
With my old laptop I had no issue to sync the mobile.  
I reinstalled git and pushed a new file to the repository with plain git and the mobile synced fine. Then I installed obsidian-git and it started to work. :)

RedHotUnicorn:
I have the same error. Laptop and iOS have merge strategy.

After trying "pull" I have message "MergeNotSupportedError"
After trying "push" I have message 
"PushRejectedError: Push rejected because it was not a simple fast-forward. Use, "force: true" to override."

nirmalhk7:
I also have the same problem: I made the changes on my laptop yesterday and opened Obsidian on my mobile today.
Seems like Obsidian ignored my changes on my laptop and simply pushed back to remote (wasn't a force push because the files I added on the laptop were shown as deleted). I tried to force push from my laptop, and voila: now when I open my phone I get a "MergeNotSupportedError".

[...]

PR #446, opened 2023-02-25
Add discard all changes command

freeman-jiang:
Resolving merge conflicts is pretty hard on mobile (#340) and though some workarounds exist, the workflow is not ideal, especially if you don't have access to your desktop. 

I think a quick solution would be to provide a `git reset --hard` command to allow mobile users to discard all their local changes to the last commit. They can then pull without conflicts and re-add their changes manually.

This PR adds a "Reset hard" command using the `gitManager`'s `discardAll` method

**Update:** I realized after an embarrassingly long amount of time that you can use the source control view to achieve the same effect, However I still think there's value in this command considering how common an operation this could be from the discussion in #340

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -383,8 +383,7 @@ export default class ObsidianGit extends Plugin {
             id: "discard-all",
             name: "CAUTION: Discard all changes",
             callback: async () => {
-                const repoExists = await this.app.vault.adapter.exists(`${this.settings.basePath}/.git`);
-                if (repoExists) {
+                if (! await this.isAllInitialized()) return false;
                     const modal = new GeneralModal({
                         options: ["NO", "YES"],
                         placeholder: "Do you want to discard all changes to tracked files? This action cannot be undone.",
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -379,6 +379,28 @@ export default class ObsidianGit extends Plugin {
             }
         });
 
+        this.addCommand({
+            id: "discard-all",
+            name: "CAUTION: Discard all changes",
+            callback: async () => {
+                const repoExists = await this.app.vault.adapter.exists(`${this.settings.basePath}/.git`);
```
Vinzent03, line 386: Instead of checking for an existing repo, better check it via.
```
if (! await this.isAllInitialized()) return false;
```

src/main.ts
```diff
@@ -1056,6 +1078,14 @@ export default class ObsidianGit extends Plugin {
         }
     }
 
+    async discardAll() {
+        await this.gitManager.discardAll({
+            dir: '/',
```
Vinzent03, line 1083: Remove the `dir` parameter here, because that's only used for directories.

Vinzent03 (changes requested):
You are right it's already possible via the sidebar, but I see the reason behind adding an extra command as well.

src/main.ts
freeman-jiang: Thanks! Good point.

freeman-jiang:
@Vinzent03 lmk if there's anything else!

Vinzent03:
LGTM, thanks for your contribution!

PR #448, opened 2023-02-27
Fix diff view issue on smartphones and tablets

### LINKED ISSUE #429: [Bug]: Diff doesn't work on iPad
### Describe the bug

When I click a changed file, the diff window opens, but it remains blank.
Title is Diff View but nothing happens. Also it locks up the window as "busy".
It did just eventually see it display "File without changes", which isn't accurate since I clicked a file with changes.

### Relevant errors (if available)

_No response_

### Steps to reproduce

Click a changed file in the Source Control tree on iPad.

### Expected Behavior

Should show a diff of the changed file.

### Addition context

_No response_

### Operating system

iOS

### Plugin version

2.16.0

tamiroh:
Fixes #429

This issue is caused by a `NotFoundError` that may occur when executing `readBlob`. The solution here is to catch the error and return `undefined` when that error occurs.

~~In addition I have simplified the logic by removing the `stagedChanged` argument and removing the branching by `stagedChanged`. In the end, all we want to see in the diff view is the difference from `HEAD`, so I think this is sufficient.~~

### DIFF
--- a/src/isomorphicGit.ts
+++ b/src/isomorphicGit.ts
@@ -758,10 +758,15 @@ export class IsomorphicGit extends GitManager {
         const stagedContent = new TextDecoder().decode(stagedBlob);
 
         if (stagedChanges) {
-            const headBlob = await readBlob({ ...this.getRepo(), filepath: filePath, oid: await this.resolveRef("HEAD") });
-            const headContent = new TextDecoder().decode(headBlob.blob);
+            const headContent: string | undefined = await this.resolveRef("HEAD")
+                .then((oid) => readBlob({ ...this.getRepo(), filepath: filePath, oid: oid }))
+                .then((headBlob) => new TextDecoder().decode(headBlob.blob))
+                .catch((err) => {
+                    if (err instanceof git.Errors.NotFoundError) return undefined;
+                    throw err;
+                });
 
-            const diff = createPatch(filePath, headContent, stagedContent);
+            const diff = createPatch(filePath, headContent ?? "", stagedContent);
             return diff;
 
         } else {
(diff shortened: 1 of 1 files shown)


Vinzent03:
No there is a difference between staged or not. You can see the unstaged changes diff and the already staged changes diff. Please bring back the feature of seeing the staged changes.

tamiroh:
@Vinzent03 

It seems I misunderstood about staging. Thank you for pointing it out!

Fixed.

Vinzent03 (changes requested):
Could you please maintain the style of using `await` in favor of `.then`?

tamiroh:
@Vinzent03 

I understand the motivation for using `await`, but in this case, I think it would be much more readable and safer to use `.then`. This is because there is exception handling, and `NotFoundError` can occur in two places: `this.resolveRef` and `readBlob`.

Without `.then`, the code would look like the followings:

```typescript
let headContent: string | undefined;
try {
    const oid = await this.resolveRef("HEAD");
    const headBlob = await readBlob({ ...this.getRepo(), filepath: filePath, oid: oid });
    headContent = new TextDecoder().decode(headBlob.blob)
} catch (err) {
    if (err instanceof git.Errors.NotFoundError) {
        headContent = undefined;
    } else {
        throw err;
    }
}
```

or

```typescript
const headContent: string | undefined = await (async () => {
    try {
        const oid = await this.resolveRef("HEAD");
        const headBlob = await readBlob({ ...this.getRepo(), filepath: filePath, oid: oid });
        return new TextDecoder().decode(headBlob.blob)
    } catch (err) {
        if (err instanceof git.Errors.NotFoundError) return undefined;
        throw err;
    }
})();
```

Vinzent03:
I see, I'm okay with using `then`. Thanks for the fix!

PR #449, opened 2023-02-28
Add test workflow

tamiroh:
Currently, there is no workflow to check the quality of code when it is pushed to the repository. However, by setting up this, we can prevent code quality degradation and achieve safer development.

First, I incorporated eslint into the workflow, which requires minimal modification of existing code. I plan to add typescript compiler checks, etc. later.


src/gitManager.ts
```diff
@@ -128,7 +128,7 @@ export abstract class GitManager {
     */
     private simplify(tree: TreeItem[]): TreeItem[] {
         for (const node of tree) {
-            while (true) {
+            for (;;) {
```
tamiroh, line 131: ESLint's no-constant-condition rule does not allow `while (true)`. Therefore, I have replaced it with `for (;;)` instead, per the following issue. If you don't like it, please disable no-constant-condition.
https://github.com/eslint/eslint/issues/5477

src/simpleGit.ts
```diff
@@ -50,9 +50,6 @@ export class SimpleGit extends GitManager {
                 process.env[key] = value;
             }
 
-            const debug = require('debug');
```
Vinzent03, line 53: That's necessary for automated debug logs. https://github.com/steveukx/git-js/blob/main/docs/DEBUG-LOGGING-GUIDE.md

src/gitManager.ts
Vinzent03: Yeah I prefer `while(true)` If disabled the rule and reverted the change. Thanks for pointing out!

src/simpleGit.ts
tamiroh: @Vinzent03 
OK👍 I added again
3723ca1

Vinzent03:
LGTM, good idea to add some automated tests!

PR #452, opened 2023-03-03
Add TypeScript compiler check

tamiroh:
Continued from #449.

TypeScript compiler check have been added into the workflow. There is no change to the existing logic. I modified several types to pass the checks, and temporarily add `@ts-ignore` to the line that needs logic change.

Please see the code differences for details.


tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "inlineSourceMap": true,
         "inlineSources": true,
         "module": "ESNext",
-        "target": "es2018",
+        "target": "ES6",
```
Vinzent03, line 7: I've downgraded that to follow the [sample plugin](https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/tsconfig.json)

package.json
```diff
@@ -17,8 +17,11 @@
         "t": ""
     },
     "devDependencies": {
+        "@codemirror/state": "^6.2.0",
+        "@codemirror/view": "^6.9.1",
         "@types/debug": "^4.1.7",
         "@types/diff": "^5.0.2",
+        "@types/hogan.js": "^3.0.1",
```
Vinzent03, line 24: What's the reason for this?

src/types.ts
```diff
@@ -1,6 +1,6 @@
 export interface ObsidianGitSettings {
     commitMessage: string;
-    autoCommitMessage: string;
+    autoCommitMessage?: string; // possibly undefined for settings migration
```
Vinzent03, line 3: I would rather make this non-undefined again as settings migration runs at the very beginning. So it's everywhere used non-undefined.

Vinzent03 (changes requested):

package.json
tamiroh: @Vinzent03 
Because without this, we will get a compile error `error TS7016: Could not find a declaration file for module 'hogan.js'.`. hogan.js is used in diff2html.

Vinzent03: Ah okay, thanks for explanation

src/types.ts
Vinzent03: I see... I think we can just omit `autoCommitMessage` in `DEFAULT_SETTINGA`.

tamiroh: I added string as default value instead of `undefined`.
1d6a92e

src/constants.ts
```diff
@@ -3,7 +3,7 @@ import { ObsidianGitSettings } from "./types";
 
 export const DEFAULT_SETTINGS: ObsidianGitSettings = {
     commitMessage: "vault backup: {{date}}",
-    autoCommitMessage: undefined, // default undefined for settings migration
+    autoCommitMessage: "vault backup: {{date}}",
```
Vinzent03, line 6: No I would rather remove the field here and use `Omit<ObsidianGitSettings,"autoCommitMessage">` as type for `DEFAULT_SETTINGS`. `autoCommitMessage` should be set to the current `commitMessage` value, in case it's undefined. That's what's done in `migrateSettings`.

tamiroh: @Vinzent03 
OK, I fixed again.
cc146da

Vinzent03:
LGTM! Thanks for the great work!

PR #463, opened 2023-03-13
feat: Add file history view

tamiroh:
## Summary

This pull request provides the feature to view file history based on Git.

This is a work in progress. If you have any feedback on this feature, please feel free to let me know!

## Usage

Execute the "Open file history" command on the file whose history you want to view.

### DIFF
--- a/src/gitManager.ts
+++ b/src/gitManager.ts
@@ -3,6 +3,7 @@ import ObsidianGit from "./main";
 import {
     BranchInfo,
     FileStatusResult,
+    GitLogResult,
     Status,
     TreeItem,
     UnstagedFile,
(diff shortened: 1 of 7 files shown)


Vinzent03:
I haven't run your code yet, so I don't know how it looks, but there is already the [Version History Diff](obsidian://show-plugin?id=obsidian-version-history-diff) plugin, which is very solid.

tamiroh:
@Vinzent03 

This is what it currently looks like (style is incomplete).

image

And I did not know about the Version History Diff. I just installed it and tried it out, and it's nice 👍 

But, the obsidian-git plugin is the first choice for using Git with Obsidian, and I think it should have some comprehensive Git-related features by itself. So it would be better to have a file history view feature, even if it is minimal.

tamiroh:
If you don't think so, feel free to close the pull request 👌

Vinzent03:
I haven't used the plugin much, but from what I see, it works very well. I see the reason for including it in this plugin as well, but it's already well implemented, so I would rather not duplicate the work. Unless you would do something very different, or you are missing important information in the other plugin's view.

tamiroh:
OK, thank you for the feedback!

PR #485, opened 2023-04-03
Enable `skipLibCheck` to fix CI

tamiroh:
Currently, the codemirror type definition file is causing compile errors and CI is failing. As an emergency workaround, I have enabled `skipLibCheck`.


Vinzent03:
Thanks for the quick fix!

PR #496, opened 2023-04-12
Enable `noUncheckedIndexedAccess` option

tamiroh:
To make development safer, I have enabled `noUncheckedIndexedAccess` option in TypeScript. This reduces the risk of runtime errors during array access.

As a result, several of the existing logic resulted in type errors, but since we need to be cautious about fundamentally fixing them, I made a workaround in this pull request by adding non-null assertions.


Vinzent03:
I don't think this is necessary as I wouldn't add an extra catch or check to the currently affected lines. I think it's fine as it is.

PR #519, opened 2023-05-28
Always pull when pull-before-push is enabled (even if nothing to commit)

m-beau:
Effectively, backups will always trigger the sequence "add > commit > pull > push", ensuring two-way sync.


.DS_Store
tamiroh, line 1: `.DS_Store` should be removed. In addition, it would be good to add it to `.gitignore`.

m-beau: Oh yes apologies

m-beau: Done.

Vinzent03:
The pull already happens a few liens above https://github.com/denolehov/obsidian-git/pull/519/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR941 So I don't see the need to pull again there?

m-beau:
This pull acts in the case where `this.settings.syncMethod != "reset"`. If `this.settings.syncMethod != "reset"`, the line you point to never gets hit. That was a real problem for my workflow - nothing was pulled from the remote in cases where I had no changes to push.

Vinzent03:
Ah thanks for explanation. I've overseen that. I've pushed a commit and changed something more. Now it's not needing the push setting at all. What do you think?

Vinzent03:
Never mind, forget what I said! There's the extra auto pull interval for that case. So it's designed that it only pulls if you push as well, and use the extra auto pull setting the rest of the time.

PR #537, opened 2023-06-17
fix: missing git file status 113

akirabaruah:
From the table of all possible combinations in https://isomorphic-git.org/docs/en/statusMatrix.


akirabaruah:
I personally hit the "113" case and was presented with a vague error: `Cannot read properties of undefined (reading '0')`.

Vinzent03:
Thanks for fixing!

akirabaruah:
Thanks for the quick review!

PR #548, opened 2023-06-30
Create README-zh-cn.md

zhiquanchi:


Vinzent03:
Hey, sorry for taking so long to comment on this pr. I'm very unsure if I want to merge this pr. I don't think it's a good idea if the README is translated, but the whole plugin messages and settings are not. And because those are always changing a bit, I don't want to translate these. So I don't see the point in only translating the README. So I will close this for now. Thanks for your work, though!

PR #553, opened 2023-07-08
fix: error caused by empty commit

XiongAmao:
If git has an empty commit, it will cause error in `history`.
I created a repository to reproduce this issue. https://github.com/XiongAmao/ob-git-plugin-issue

Snipaste_2023-07-08_13-10-19


Vinzent03:
Thanks for your fix, but I committed it now directly to the master branch (257147311cf65a2b5dedf957f4c71ad9624ce7be), because I would use different syntax and proper formatting. Thanks for finding the issue though!

PR #569, opened 2023-07-30
Adds .obsidian/workspace-mobile.json to ignore.

Supetorus:
I noticed this file was generated since I use obsidian on my iPhone, so it should probably be ignored too if you don't want workspace data.


Vinzent03:
Great idea! Thanks for the addition.

Issue #141, opened 2021-11-24
[Bug]: custom Git binary path using wsl doesn't work


229c9cf0:
[note: suspected cause at the very end; skip ahead and come back if that's not it]

I have Git installed in WSL and added a shell script in my home that will take care of connecting to the `ssh-agent` etc. followed by actually running the requested git command. Despite the command working in both PowerShell and plain CMD, Obsidian Git reports "cannot run git command".

 *  Are you sure that the custom git binary path code actually works?
 *  If yes, what's the required format?

I have tried both `C:\Windows\System32\wsl.exe /home/REDACTED/bin/ogit` as well as just `wsl.exe /home/REDACTED/bin/ogit` or `wsl /home/REDACTED/bin/ogit` (in case there's a problem with handling backslashes...)

In CMD and PowerShell, both work:

```
C:\Users\REDACTED\Desktop\maps\obsidian>wsl /home/REDACTED/bin/ogit status
On branch master
Your branch is ahead of 'origin/master' by 2 commits.
  (use "git push" to publish your local commits)

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        modified:   .obsidian/plugins/obsidian-git/data.json


C:\Users\REDACTED\Desktop\maps\obsidian>C:\Windows\System32\wsl.exe /home/REDACTED/bin/ogit status
On branch master
Your branch is ahead of 'origin/master' by 2 commits.
  (use "git push" to publish your local commits)

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        modified:   .obsidian/plugins/obsidian-git/data.json

C:\Users\REDACTED\Desktop\maps\obsidian>C:\Windows\System32\wsl.exe /home/REDACTED/bin/ogit push
Host key fingerprint is SHA256:<redacted>
Enumerating objects: 38, done.
Counting objects: 100% (38/38), done.
Delta compression using up to 8 threads
Compressing objects: 100% (27/27), done.
Writing objects: 100% (29/29), 6.13 KiB | 184.00 KiB/s, done.
Total 29 (delta 4), reused 0 (delta 0)
To <redacted>:repos/obsidian-vault.git
   <redacted>..<redacted>  master -> master
```

In Obsidian, all variants fail.

So there seems to be some special formatting / adjustment to the command required to make it work when run from Obsidian. (Or maybe it just doesn't work at all?)

-----

[...]

PR #587, opened 2023-08-31
Adjust git.cwd to use a relative path to git root

H3mul:
This is a simple adjustment for the git cwd command to use a relative path to the git root instead of an absolute path. It should keep the behavior identical, but it solves an issue I ran into in my setup (described below).

More details on `--show-cdup` vs `--show-toplevel` here:
https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---show-cdup

Motivation:
I ran into this issue during my esoteric obsidian-git setup using WSL2 (following the solution in #141). Basically, I want to use WSL's git instead of native windows `git.exe`, since I set it up with `git-crypt` and it has my ssh credentials. I ran into the issue that in this case, `git.revparse("--show-toplevel")` will return the root as an absolute WSL path (eg, `/mnt/c/Users/...`). `git.cwd()` checks path existence prior to actually switching the working directory, but does so using `fs` - which will be using the windows filesystem instead, and, of course, not find the root directory, and throw an error.

A quick solution is to pass a relative path to `git.cwd()` instead of an absolute path, as those will work the same way in WSL and Windows filesystems for a given repo. This is exactly what the `--show-cdup` flag returns. Note that it will be an empty string if we are already at the repo root, hence the additional existence check.

Some sources for my deep dive:

https://github.com/steveukx/git-js/blob/d64b31ca8670edd7af5a7fe5658516f5717c79a8/simple-git/src/lib/tasks/change-working-directory.ts#L7
https://github.com/steveukx/git-js/blob/d64b31ca8670edd7af5a7fe5658516f5717c79a8/simple-git/src/lib/utils/util.ts#L74
https://github.com/kwsites/file-exists/blob/13f0789cbd8df52588acdd185cd33de3b77fb342/src/index.ts#L10C13-L10C18

### DIFF
--- a/src/gitManager/simpleGit.ts
+++ b/src/gitManager/simpleGit.ts
@@ -2,7 +2,7 @@ import { spawnSync } from "child_process";
 import debug from "debug";
 import { FileSystemAdapter, normalizePath, Notice } from "obsidian";
 import * as path from "path";
-import { sep } from "path";
+import { sep, resolve } from "path";
 import simpleGit, * as simple from "simple-git";
 import { GIT_LINE_AUTHORING_MOVEMENT_DETECTION_MINIMAL_LENGTH } from "src/constants";
 import { LineAuthorFollowMovement } from "src/lineAuthor/model";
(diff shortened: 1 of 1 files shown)


Vinzent03:
Looks really promising! Thanks for that fix. Will test later or start next week. Please ignore all tests other than the format test. I've messed them up in my latest commits. But the format seems actually to be wrong.

Vinzent03:
Have you tested this on normal windows without wsl? I'm getting the following error:
```
not a git repository (or any of the parent directores): .git
```
So I don't think the relative cwd is working properly.

H3mul:
Interesting, I havent, let me do some testing as well. I would expect that in this case, `git.checkIsRepo()` to also fail.

H3mul:
Looks like you were right, the relative path cwd seems to break. Adjusted this feature to resolve the relative path to an absolute one in the nodeJS filesystem, which works both in my WSL case and the windows `git.exe` case based on my testing

H3mul:
Apologies for the force push, rebased the changes from this issue upon my fork's master to be up-to-date and forgot this issue tracks that branch. No new changes have been introduced since my last comment.

PR #602, opened 2023-09-10
Fix CI errors

tamiroh:
To fix the CI errors, I mainly updated the svelte config in `.eslintrc`.

See comments for other fixes and additional information.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,7 @@
         "eslint:recommended",
         "plugin:@typescript-eslint/eslint-recommended",
         "plugin:@typescript-eslint/recommended",
-        "plugin:svelte/recommended"
+        "plugin:svelte/base"
     ],
     "parserOptions": {
         "sourceType": "module",
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,28 +5,34 @@
         "node": true
     },
     "plugins": [
-        "@typescript-eslint",
-        "svelte3"
+        "@typescript-eslint"
     ],
     "overrides": [
         {
             "files": [
                 "*.svelte"
             ],
-            "processor": "svelte3/svelte3"
+            "parser": "svelte-eslint-parser",
+            "parserOptions": {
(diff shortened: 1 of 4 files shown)


.eslintrc
```diff
@@ -5,28 +5,37 @@
...
             "files": [
                 "*.svelte"
             ],
-            "processor": "svelte3/svelte3"
+            "parser": "svelte-eslint-parser",
+            "parserOptions": {
+                "parser": "@typescript-eslint/parser"
+            },
+            "rules": {
+                "no-undef": "off" // TODO: remove this line
```
tamiroh, line 20: Here I temporarily suppressing the `no-undef` errors, which currently occurs for global objects such as `window`, `app` and so on. I can't solve this problem right now, so this is a temporary work around.

src/ui/sourceControl/components/stagedFileComponent.svelte
```diff
@@ -11,7 +11,6 @@
     export let view: GitView;
     export let manager: GitManager;
     let buttons: HTMLElement[] = [];
-    $: formattedPath = change.vault_path;
```
tamiroh, line 14: I removed this line to suppress the following error.

```
14:8   error  'formattedPath' is defined but never used  @typescript-eslint/no-unused-vars
```

This variable is not actually referenced and should not affect the behavior.

src/main.ts
```diff
@@ -1251,7 +1251,7 @@ export default class ObsidianGit extends Plugin {
             this.app.workspace.iterateAllLeaves((leaf) => {
                 if (
                     leaf.view instanceof MarkdownView &&
-                    leaf.view.file.path == file.path
+                    (leaf.view.file as any).path == file.path
```
Vinzent03, line 1254:
```suggestion
                    leaf.view.file?.path == file.path
```

Vinzent03 (changes requested):

Vinzent03:
Thanks for the fixes!

PR #607, opened 2023-09-22
feat: add setting for maximum number of files to display

Nikolai2038:
Hi!

I was working with Obsidian Git in big repository with a lot of notes. Got `Too many changes to display` error too often and was forced to use external VCS tool. Then I looked at the source code of this plugin and find out, that the limit is `500` and there is no setting to change that. So I add new setting `maximumNumberOfFilesToDisplay` with default value `500`. Hope I did it right!

Also, I noticed that when the value of this new setting is reduced to the error `Too many files changes` (when `refresh()` is called), then an error in console `Uncaught (in promise) TypeError: Cannot read properties of null (reading 'firstElementChild')` occurs: 

image

So I add `if (!btn) { return; }` for that to `refresh()` method in `sourceControl.svelte`.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -30,6 +30,7 @@ export const DEFAULT_SETTINGS: Omit<ObsidianGitSettings, "autoCommitMessage"> =
         changedFilesInStatusBar: false,
         showedMobileNotice: false,
         refreshSourceControlTimer: 7000,
+        maximumNumberOfFilesToDisplay: 500,
         showBranchStatusBar: true,
         setLastSaveToLastCommit: false,
         submoduleRecurseCheckout: false,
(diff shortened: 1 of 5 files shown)


Vinzent03:
I want to keep the plugin more simple as there are already many settings. So I won't add this setting. I don't think many people encounter that file limt. What are you doing that you encounter this? You should still be able to commit and push via the command palette.

Nikolai2038:
Yeah, that's why I use external VCS to work with GIT instead of Obsidian Git. You are right - I can still commit and push via this plugin, but I don't quite understand, why you need to limit it's functionality and don't provide the way to customize it.

Quite sad, but it is your choice.

Vinzent03:
If I remember correctly it was for performance reasons. The view became unresponsive with too many files listed. There may be a better way to solve this, but that was the easiest one at that point.

PR #624, opened 2023-10-15
Simplify expression with `!`

tamiroh:
Since `leaf.view.state.staged` is of type boolean, the logical negation operator `!` simplifies the expression.


Vinzent03:
I would rather keep it as it is, because I'm not entirely sure if that's always a boolean. It doesn't hurt to keep it.

PR #656, opened 2023-12-16
fix: update settings grammar

PScoriae:
What my changes do:
- add `formatMinutes` function
  - use "minute" when autoSaveInterval is 1
  - else use "minutes"
  - Prevents the scenario phrase "1 minutes."
- Improve grammar for `autoBackupAfterFileChange` plugin
  - turn message into multiline string for easier readability in code
 
I did think about omitting the interval value if it's set to 1 i.e.:
- "Every minute."
- "Every x minutes."

But I believe being more explicit with whatever interval value the user has set reinforces the idea that the application has registered their input.

### DIFF
--- a/src/setting/settings.ts
+++ b/src/setting/settings.ts
@@ -23,7 +23,7 @@ import {
 } from "src/lineAuthor/model";
 import ObsidianGit from "src/main";
 import { ObsidianGitSettings, SyncMethod } from "src/types";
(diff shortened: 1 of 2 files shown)


antonkesy (approved):
LGTM

Vinzent03:
Thanks for your contribution!

PR #657, opened 2023-12-16
refactor: Use boolean values for `getConflict()` and `setConflict()`

injust:
Refactors `getConflict()` and `setConflict()` to use boolean values instead of strings, similar to `getPluginDisabled()` and `setPluginDisabled()`.


Vinzent03:
Thanks for your contribution!

PR #674, opened 2024-01-15
Update settings.ts

qfletcher89:
Added momentjs specification for more {{date}} formats.

Hey 👋 didn't know which formats to use for the date until i took a look at the source code. Hopefully this will help others


Vinzent03:
That's a great point, makes totally sense to add a link there. Though, making the link clickable would be even better. In some examples in the same file I'm using `descEl.innerHTML`, but I think there was a recommendation to not use `innerHTML`.
If you don't know how to accomplish this, I will try next week.

PR #676, opened 2024-01-15
Added setting to disable 'No changes...' popups

### LINKED ISSUE #675: Feature Request: Setting to hide popups when there are no changes.
I propose a new setting that is displayed under the "Disable notifications" setting that only appears when notifications are enabled. The setting would simply hide the "No changes..." popups but keep all other notifications.

**Reason for the Feature:**
I like to have notifications turned on while I work to see when my changes were backed up which I have set for every 10 minutes. I also like to keep Obsidian open off to the side when I'm not using it, but I find it distracting to have the popup notification display every 10 minutes when nothing has changed.

jtams:
When notifications aren't disabled this new setting will be displayed. Enabling this setting, will prevent the "No changes to push" and "No changes to commit" messages from displaying. This doesn't prevent
these notifications from appearing in the status bar.

Adds feature to resolve #675

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1728,8 +1728,10 @@ I strongly recommend to use "Source mode" for viewing the conflicted files. For 
         this.statusBar?.displayMessage(message.toLowerCase(), timeout);
 
         if (!this.settings.disablePopups) {
-            if (!this.settings.disablePopupsForNoChanges
-                || !message.startsWith("No changes")) {
+            if (
+                !this.settings.disablePopupsForNoChanges ||
+                !message.startsWith("No changes")
+            ) {
                 new Notice(message, 5 * 1000);
             }
         }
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -1728,7 +1728,11 @@ I strongly recommend to use "Source mode" for viewing the conflicted files. For
         this.statusBar?.displayMessage(message.toLowerCase(), timeout);
 
         if (!this.settings.disablePopups) {
-            new Notice(message, 5 * 1000);
+            if (!this.settings.disablePopupsForNoChanges
+                || (this.settings.disablePopupsForNoChanges
+                && !message.startsWith("No changes"))) {
```
Vinzent03, line 1733:
```suggestion
                || !message.startsWith("No changes")) {
```
`disablePopupsForNoChanges` should always be true if the first case isn't met.

Vinzent03 (changes requested):
Thanks for your contribution! Sounds like a great idea. Just left a small comment.

PR #684, opened 2024-01-28
Prevent text superposing git blame

cesarmontalverne:
Change to prevent git blame gutter from superposing text in smaller screens. There is no change for screens larger than 1000px
Before:
image

After:
image

Before and After for larger screens:
image

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -511,7 +511,7 @@
     white-space: pre; /* Keep spaces and do not collapse them. */
 }
 
-@media(max-width:1000px){
+@media(max-width:800px){
     /* hide git blame gutter not to superpose text */
     .cm-gutterElement.obs-git-blame-gutter {
         display: none;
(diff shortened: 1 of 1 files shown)


Vinzent03:
I don't have that issue when shrinking the window size. Do you use any theme or custom css snippets?

cesarmontalverne:
It's a custom css snippet that hides git blame if the window is smaller than 1000px in width. It's like 5 lines. No themes.

When I divide my screen to multitask, git blame superposes my text. I just came up with a small snippet to prevent that. 

I love this plugin by the way - thanks for having made it!

Vinzent03:
The real issue here is the minimal theme, which somehow occupies the gutter, which the line authoring uses. It happens in full screen as well, with readable line length disabled.

I will merge this at a width of 800px, but just because it's a general fix to hide it on small widths.

styles.css
```diff
@@ -510,3 +510,10 @@
     padding: 0px 6px 0px 6px;
     white-space: pre; /* Keep spaces and do not collapse them. */
 }
+
+@media(max-width:1000px){
```
Vinzent03, line 514:
```suggestion
@media(max-width:800px){
```

Issue #609, opened 2023-09-25
[Bug]: Pushing broken on iphone obsidian git


AdityaKrishnaProjects:
### Describe the bug

Pushing stops working on iphone obsidian git, but works briefly after recloning. 

### Relevant errors (if available) from notifications or console (`CTRL+SHIFT+I`)

```shell
No errors from console. Phone gets stuck on initializing push. 

https://media.discordapp.net/attachments/232554307755573248/1155686950875959326/image.png?width=459&height=993
```

### Steps to reproduce

Clone a repo using alpine client (iphone app iSH). 

Set author information and provide authentication key. 

Use obsidian git, and eventually push will break. Pull and commit fully functional. 

### Expected Behavior

Should be able to push. 

### Addition context

This issue arises on the phone after 1-2 days of using obsidian git as my sync method. I am using iSH as the client through which I am initially cloning the repo, but then from there using obsidian git to manage my push, pull and commits. After a fresh clone the issue stops for a bit, but eventually return. My guess is that the request somehow times out. My confusion is that it somehow is fixed when I clone the repo fresh. 

I have a suspicion that this may be related to iSH, but I do not know how to investigate this further as I get no error message. 

### Operating system

iOS

### Installation Method

None

### Plugin version

Newest

AdityaKrishnaProjects:
Hasn't been an issue for a couple days with current clone of vault, which leads me to believe it's not an issue with vault size

idalsin:
I've been seeing this as well on iOS, though unsure how to consistently reproduce. After editing a file I will commit all changes, then push - the "initializing push" banner message appears, hangs for approx 60-75 seconds, and then it looks like the plugin restarts. The push does not complete and there is no error message. This happens on both manual push and the scheduled backup push.

kunfoo:
I leave this as a note for anyone who has the same problem: for some reason, after restarting the iPhone, push works again.

AdityaKrishnaProjects:
Just wanted to provide an update that this error still occurs for me on mobile.

nickdavila:
Same error occurs for me as well!

dedouwe26:
Hello, i have the same issue on my end.

Vaibess:
Has anyone found a fix for this? I’m experiencing a very similar issue but I’ve never been able to push from IOS, though I am only a week into this. PC git sync has been flawless otherwise.

dv-markov:
[...]

Issue #574, opened 2023-08-05
[Bug]: Unable to push from iOS app


dv-markov:
### Describe the bug

The repo is cloned from remote, authentication set with personal access token (classic), the app is working, all fine. 
After couple of months of use there is an error appearing when I try to push changes to remote.
Changes are added, commited, but when I try to push, get the following error:
HttpError: HTTP Error: 500 500

### Relevant errors (if available) from notifications or console (`CTRL+SHIFT+I`)

```shell
HttpError: HTTP Error: 500 500
```

### Steps to reproduce

Make changes or create new file, add, commit, push

### Expected Behavior

Changes are successfully pushed to remote

### Addition context

The problem can be solved only by deleting the local repo from the device and cloning it from remote all over again. However I expect that is not should be happen every one or two months.

### Operating system

iOS

### Installation Method

None

### Plugin version

2.19.1

Vinzent03:
I can't do much here, because I don't know what the issue is. The 500 error code means that an error happened on the server you are pushing to. Are you using a custom git hosting server? Either way, I don't know what exactly is wrong and can't reproduce.

michaelgoldenn:
In case you're still having this problem, or anyone else comes along with the same problem, it could be your version of IOS. I would get this error consistently a couple hours after re-cloning my vault. Once I updated to the latest IOS version the problem vanished.

Issue #653, opened 2023-12-15
[Bug]: Push doesnt work on mobile


MakiRollDev:
### Describe the bug

So I had to copy my repository onto my phone by hand (via USB on my PC). 
The problem is that when I try to push my changes to Github, it says "MissingParameterError: The function requires a 'remote or URL''"

I have tried:
- creating a new remote on my phone
- switching to remote branch
- appending '.git' to the repo URL
- 

### Steps to reproduce

I do not have the hardware to test this on other devices
- Set up obsidian git on desktop
- Copy it over to your android device and open it in obsidian
- set up authentication to github (Name and Token)
- commit all changes
- push

### Operating system

Android

### Installation Method

None

### Plugin version

2.22.0

kaisersakhi:
@MakiRollDev Did you find any workaround?

MakiRollDev:
@kaisersakhi No sadly not I just switched to Google drive

trinityzhub:
Same error for me in my  Android Phone 
```
MissingParameterError : The Function requires a "remote OR url" parameter . But none was provided
```

Irvus:
same problem. I would like to give some logs or more info, but don't know how

Irvus:
no, i no idea how to open dev tools from android. ok, android 11 rkq1.200826.002, MIUI 12.5.6, pull works correctly. field "username on your git server does no effect, i changed it to completly incorrect, and all worked after reboot of the phone. Also, the new token i generated in github was never used. There are files that were modifed after last commin inside .git of the phone, some /objects, index, /refs/heads and so on

Also I add .git/config files from windows and android. As I see, username and psw fields do nothing because url was not modified after copying from win, but author name and email were changed. Maybe that's okay for git security, i'm bad in it
[config-windows.txt](https://github.com/user-attachments/files/17941283/config-windows.txt)
[config-android.txt](https://github.com/user-attachments/files/17941284/config-android.txt)


Now something I was doing: i used [this guide](https://www.reddit.com/r/ObsidianMD/comments/17odzjb/obsidian_android_syncing_via_github_in_2023/) to sync my devices

>  Step 5: Add the folder as a vault in Obsidian
> 
>     1. Open Obsidian
> 
>     2. Tap the sidebar icon in the top left corner
> 
>     3. Click the vault name in the top left corner, click "Manage valuts..."
> 
>     4. Click "Open folder as vault" and navigate the the folder you copied over
> 
>     5. Obsidian Git should try to fetch
[...]

PR #687, opened 2024-01-29
fix(mobile): resolve push hanging

ras0q:
ref: https://github.com/denolehov/obsidian-git/issues/609 https://github.com/denolehov/obsidian-git/issues/574

`async push()` for mobile is implemented using `this.branchInfo()`, but this branch info should be converted to latest commit hash.

### DIFF
--- a/src/gitManager/isomorphicGit.ts
+++ b/src/gitManager/isomorphicGit.ts
@@ -489,8 +489,16 @@ export class IsomorphicGit extends GitManager {
             const status = await this.branchInfo();
             const trackingBranch = status.tracking;
             const currentBranch = status.current;
+
+            if (trackingBranch == null || currentBranch == null) {
+                return 0;
+            }
+
+            const localCommit = await this.resolveRef(currentBranch);
+            const upstreamCommit = await this.resolveRef(trackingBranch);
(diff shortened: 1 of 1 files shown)


Vinzent03:
Did you encounter one of these two branches being null? I'm pretty sure the ref for a Tree can also just be the branch name.

ras0q:
I don't encounter this and using both the branch name and the commit hash can be the ref for a tree.
However by using the commit hash, `git.TREE({ ref })` can be resolved more quickly.
https://github.com/isomorphic-git/isomorphic-git/blob/53310498e9dd84a213197de2d69047ce733580e6/src/managers/GitRefManager.js#L193

ras0q:
Here is my brief benchmark!
https://github.com/ras0q/isomorphic-git-walk-test

ras0q:
@denolehov @Vinzent03 could you review this PR? Thanks!

antgiant:
Adding #653 as another instance.

I’m experiencing this too

Vinzent03:
When digging the code of `TREE()` it uses `resolveRef()', just like you do now manually. I don't see what this pr actually changes.

PR #722, opened 2024-04-15
fix: getRelativeRepoPath for the case of vault-inside-repo

### LINKED ISSUE #660: [Bug]: Line author fails when the vault is a subdir of the git repo
### Describe the bug

For vaults that are not the root of a git repository the "line author" feature doesn't work

```
obsidian-git. hash-object failed: fatal: could not open 'Readme.md' for reading: No such file or directory
eval @ plugin:obsidian-git:29800

plugin:obsidian-git:29119 Uncaught (in promise) Error: fatal: could not open 'Readme.md' for reading: No such file or directory

    at Object.action (plugin:obsidian-git:29119:25)
    at PluginStore.exec (plugin:obsidian-git:29144:25)
    at eval (plugin:obsidian-git:26649:43)
    at new Promise (<anonymous>)
    at GitExecutorChain.handleTaskData (plugin:obsidian-git:26647:16)
    at GitExecutorChain.eval (plugin:obsidian-git:26631:44)
    at Generator.next (<anonymous>)
    at fulfilled (plugin:obsidian-git:25476:24)
```

### Steps to reproduce

- Create a new git repo `bob`
- Create a new vault as a subdir in the repo `bob/vault`
- Add some notes, commit, and turn on "line author"

### Expected Behavior

The rest of obsidian git works fine for vaults that are not the root of a repository, so I would expect this feature to also work.

### Addition context

_No response_

### Operating system

macOS

### Installation Method

None

### Plugin version

2.22.0

tomoyanonymous:
This fixes #660.

Currently, `getRelativeRepoPath` considers only the case of git repository is located inside the vault.
If the vault is inside a subdirectory of git repository (this is a common case when you are using a self-host publishing system like [Quartz](https://github.com/jackyzha0/quartz)), line authoring feature won't work.

This code resolves the relative path between git repo and vault using `path` library.

### DIFF
--- a/src/gitManager/gitManager.ts
+++ b/src/gitManager/gitManager.ts
@@ -1,4 +1,4 @@
-import { App, FileSystemAdapter } from "obsidian";
+import { App, FileSystemAdapter, Platform } from "obsidian";
 import ObsidianGit from "../main";
 import * as path from 'path';
 import {
@@ -124,6 +124,10 @@ export abstract class GitManager {
     // @param doConversion - If false, the path is returned as is. This is added because that parameter is often passed on to functions where this method is called.
     getRelativeRepoPath(path_str: string, doConversion: boolean = true): string {
         if (doConversion) {
+            if (Platform.isMobileApp && this.plugin.settings.basePath.length > 0) {
+                //Expect the case that the git repository is located inside the vault on mobile platform currently.
+                return path_str.substring(this.plugin.settings.basePath.length + 1);
+            }
             let adapter = this.plugin.app.vault.adapter;
             if (adapter instanceof FileSystemAdapter) {
                 const folder = adapter.getBasePath();
--- a/src/gitManager/gitManager.ts
+++ b/src/gitManager/gitManager.ts
@@ -1,6 +1,6 @@
 import { App, FileSystemAdapter, Platform } from "obsidian";
 import ObsidianGit from "../main";
-import * as path from 'path';
+import * as path from "path";
 import {
     BranchInfo,
     DiffFile,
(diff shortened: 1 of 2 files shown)


src/gitManager/gitManager.ts
```diff
@@ -123,8 +124,12 @@ export abstract class GitManager {
     // @param doConversion - If false, the path is returned as is. This is added because that parameter is often passed on to functions where this method is called.
     getRelativeRepoPath(path: string, doConversion: boolean = true): string {
         if (doConversion) {
-            if (this.plugin.settings.basePath.length > 0) {
-                return path.substring(this.plugin.settings.basePath.length + 1);
+            let adapter = this.plugin.app.vault.adapter;
+            if (adapter instanceof FileSystemAdapter) {
```
Vinzent03, line 128: On mobile, this method does nothing now. I think we should fall back to the old implementation on mobile.

src/gitManager/gitManager.ts
```diff
@@ -1,5 +1,6 @@
-import { App } from "obsidian";
+import { App, FileSystemAdapter } from "obsidian";
 import ObsidianGit from "../main";
+import { relative, join } from 'path';
```
Vinzent03, line 3: please import `path` with `as path` here to make the usage from the path library more clearly.

Vinzent03 (changes requested):

src/gitManager/gitManager.ts
tomoyanonymous: Does it mean `import * as path from 'path';` ?

Vinzent03:
Hey, thanks for this pr and your contribution! This was a great start, but I changed it so that the basePath setting is not needed to point to the repo root, but can be a subdirectory. Additionally, I needed to align the `getRelativeVaultPath` method as well. Let me know if you think something is off now. Thanks again.

Vinzent03:
Okay great. But the submodule issue was there before this pr and persisted after your version of this pr as well, right?

tomoyanonymous:
Yes.

0atman:
thanks folks!

PR #745, opened 2024-07-02
Introduce testing with Vitest

tamiroh:
This pull request sets up an environment for testing with Vitest ⚡. In addition, as a first step, I have added some tests for utils.

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,0 +1,10 @@
+import { defineConfig } from "vitest/config";
+import path from "path";
+
+export default defineConfig({
+    test: {
+        alias: {
+            obsidian: path.resolve(__dirname, "mocks/obsidian.ts"),
+        },
+    },
+});
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,3 +58,14 @@ jobs:
               run: pnpm install
             - name: Run build
               run: pnpm run build
+    test:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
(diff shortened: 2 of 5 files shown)


vitest.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "vitest/config";
+import path from "path";
+
+export default defineConfig({
+    test: {
+        alias: {
+            obsidian: path.resolve(__dirname, "mocks/obsidian.ts"),
```
tamiroh, line 7: `obsidian` needs to be mocked. The functions I have tested in this PR does not depend on `obsidian`, so it is currently an empty file.

See: https://github.com/vitest-dev/vitest/issues/4029#issuecomment-1837183186

Vinzent03:
Thanks for your effort to add tests to this plugin. However, I would greatly appreciate, if you would open an issue to discuss such huge decisions beforehand. I've thought a lot about adding tests, but it's not that easy, as many parts are highly integrated with Obsidian or the actual git cli. Testing some independent helper methods like you did is probably still a great idea.

I haven't worked much with javascript testing frameworks, but from looking at other plugins, I would rather use jest instead of vitest, because jest is more popular. From what I quickly read, using vitest makes more sense if we would use vite as well.

tamiroh:
OK, this was a bit of an sudden proposal, and I didn't explain it well enough. I am closing this pull request for now and will open an Issue again. Thank you for your review!

As for Jest, Jest is certainly the more mature choice, but I think Vitest works better with ESM & TypeScript, and the API is compatible with Jest, so if you don't like it, it's actually not difficult to switch to Jest.

PR #748, opened 2024-07-05
replace git-pull-request icon with lucide icon

nathonius:
The current icons for the diff and source control view don't have the proper `stroke` and `fill` config, so they end up always being black. This replaces them with the built-in lucide icon for git pull requests.

Before:
image

After:
image

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -60,7 +60,7 @@ export const DEFAULT_SETTINGS: Omit<ObsidianGitSettings, "autoCommitMessage"> =
 export const SOURCE_CONTROL_VIEW_CONFIG = {
     type: "git-view",
     name: "Source Control",
-    icon: "git-pull-request",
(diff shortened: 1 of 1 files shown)


Vinzent03:
The current icons work fine for me. I'm pretty sure they don't have to be prefixed with `lucide` from reading the docs. I'm nowhere defining the icon myself, so they already have to be coming from the included lucide icons.
I think you are using some custom plugin or theme, which break them. Your current pull request icon doesn't match the actual from lucide and how it's looking for me, because you already have the arrow in it.
image


https://docs.obsidian.md/Plugins/User+interface/Icons

nathonius:
...huh. You're right, it looks fine in a fresh vault. After some debugging, turns out something in the Jira integration plugin is overriding or conflicting. I'll figure out what's causing the conflict on that side - that plugin should probably not be replacing the icon.

PR #781, opened 2024-10-04
Made max file count display settings configureable

ugurozturk:


Vinzent03:
What's the reason for adding this option? How often do you have more than 500 changes and still work with the source control view. I don't want to clutter the settings page too much.

ugurozturk:
There are two reasons for this:

I was a Notion user, and when I moved to Obsidian, I had to reorganize most of my folders. It wasn’t something I could do all at once because you realize things gradually as you use Obsidian.

As a multi-device user (4 devices), I only use Git on my main device because I frequently encountered conflicts when using the others. Obsidian's Git integration and iCloud backups were getting mixed up. I use Git as a snapshot-like backup system, so I only perform Git operations on the main device.

As a result, when I run Git once a week, there are usually a lot of changes to commit.

ugurozturk:
Thanks for the new implement https://github.com/Vinzent03/obsidian-git/commit/fe1aedb0a12d4e0b5d4a81821a8d08c1417dd6b8, i am closing this

Vinzent03:
Yeah I thought about approaching the prevention of lags due to too many elements in the dom, differently than before. It's not released yet, and I will probably change minor things to not always clone the whole list. Have you tried it, and if so how did it work for you?

PR #805, opened 2024-11-08
fix: Make the spelling of .gitignore consistent

byjadams:
Everywhere else in the commands and the docs do you use `.gitignore` except for this one command. Just making it consistent.


Vinzent03:
Thanks for your contribution!

PR #813, opened 2024-11-22
Minor typo fix (.gitignroe)

tylere:


Vinzent03:
Thanks for the quick fix!

PR #834, opened 2025-01-06
Proposed solution to #746 : Allow disabling the git-commit-routine via hotkey-toggle

Gewerd-Strauss:
Hello,

this is a proposed solution to the issue outlined in #746, allowing the user to temporarily suppress the plugin committing files via a hotkey-toggleable config-setting.
I remain that I have not found any issues arising from these changes on my end. I've tried to implement the solution comparable to existing code.

I am open for suggestions/review.

Thank you.
Sincerely,
~Gw


Gewerd-Strauss:
I have since resolved the merge conflicts, but I am not entirely sure if they should be resolved this way. I'd leave this up to maintainers, and can revert these changes for manual review.

src/constants.ts
```diff
@@ -30,6 +30,7 @@ export const DEFAULT_SETTINGS: Omit<ObsidianGitSettings, "autoCommitMessage"> =
         refreshSourceControl: Platform.isDesktopApp,
         basePath: "",
         differentIntervalCommitAndPush: false,
+        globalEnable: true,
```
Vinzent03, line 33: I would rename this to `temporaryDisableAutomatics`
So the toggle gets a new meaning as well. I think it's better to add a toggle to just disable all automatics all together. So push and pull as well.

src/main.ts
```diff
@@ -1216,11 +1218,7 @@ export default class ObsidianGit extends Plugin {
 I strongly recommend to use "Source mode" for viewing the conflicted files. For simple conflicts, in each file listed above replace every occurrence of the following text blocks with the desired text.
 
 \`\`\`diff
-<<<<<<< HEAD
```
Vinzent03, line 1219: This should not be deleted, it may look like a merge conflict, but is meant to be this way.

src/setting/settings.ts
```diff
@@ -64,6 +64,22 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
 
         let setting: Setting;
         if (gitReady) {
+            new Setting(containerEl).setName("Activation").setHeading();
```
Vinzent03, line 67: I don't think a setting is needed for this. (I want to keep the amount of settings rather low) I guess a command is enough.

src/main.ts
```diff
@@ -785,6 +785,8 @@ export default class ObsidianGit extends Plugin {
         amend?: boolean;
     }): Promise<boolean> {
         if (!(await this.isAllInitialized())) return false;
+        if (!(this.settings.globalEnable)) return false; // if globalEnable-checkbox is disabled, return without committing
```
Vinzent03, line 788: I think you should still be able to manually commit via Obsidian. Only the automatic commits should be disabled, don't they? So I would not add that check here, but reset the automatics and don't start them.

Vinzent03 (changes requested):

Gewerd-Strauss (commented):
Note that I am not sure if this callback must be asynchronous or not, this is not my specialty.

Vinzent03:
I've changed quite a bit of stuff now. It's no longer persisted to settings, but the current session only.

Gewerd-Strauss:
Hmm.


> Does this still fit your needs and works how you would expect?

I am not sure.

On the one hand this is a good idea, because if the user relies on the plugin for backup-purposes they won't all-too-bad sudden realisations of 'oh, I haven't committed in ten days'. 

What I am not sure about is if this new method  would also suppress the execution of calls to git when the boolean is changed while an async to asny commit() is scheduled (aka: does an async routine respect object updates issued after the routine's async call was set up? I don't know typscript's intrinsics to that point.)


A good thing about the initial design was that if I disabled the plugin by a config-setting which was queried during commit():


```ts
if (!(this.settings.temporaryDisableAutomatics)) return false; // if temporaryDisableAutomatics-checkbox is disabled, return without committing
```


it would persist over resets. There are basically two use-cases for this functionality (at least on my end):

- temporarily disable automatic commit/push/pull while working in f.e. VS Code/wanting to curate commits myself/..., and re-enable it afterwards. Since obsidian's wealth of plugins can sometimes hinder work in other programs if they simultaneously operate on the same 'vault' (think stuff like `auto-note-mover`, linter, probably a couple other as well), there can still exist reasons beyond `obsidian-git` that might cause the user to temporarily close Obsidian. I feel like then re-opening obsidian and git coming back to life might be a source of frustration.
- have automatics disabled across obsidian restarts allows me to use obsidian even after closing it, without having to worry about suddenly re-disabling it. Essentially, this made it an explicit master toggle, so you knew that if you disabled it, you also had to enable it _again_. 







---

From a design perspective I feel like this explicit manner doesn't mix well with a 50%-automation? (Please excuse me for this probably clunky explanation, I don't really know how to voice this effectively for some reason. I hope you get the gist.)

In principal I like the solution you propose, I just don't like the mix-up it implies.

---

A potential middle-ground between both could be the implementation of some grace-period and a warning to the user, enacted/issued upon restart?
Such that when opening Obsidian while the last session was closed with automatics inhibited, the user would get some notice `Last session was closed with 'obsidian-git' disabled. Automatic routines will resume in X seconds, unless automatics are disabled <by hotkey/methodX/methodY/methodZ>  beforehand.`.

That would resolve the danger of users forgetting their git routines are not working (and thus, they are not backing up/syncing/...), and allow pseudo-persistent deactivation across app-restarts.

And if the use re-enables automatics before closing, said notices would obviously not trigger. 

---

Does this make sense?

Gewerd-Strauss:
As I said, I am unsure if the changes you propose make the solution better, or not. In parts because I am not that well-versed in typescript's inner workings and thus have a harder time judging it.

Could you provide feedback on these thoughts/the middle-ground I propose?

Vinzent03:
Sorry for taking so long to coming back to this pr!
What do you think of the following changes to the current state:
- Paused automatics are indeed persisted over Obsidian restarts by storing it in the localstorage
- When starting Obsidian/the plugin and the automatics are paused a notification is shown to warn the user that their work is not auto commited. But not with some timeout as you proposed.

To your question about the async commits. This would not stop commits that are already in the queue or already running. But I don't think it should be the goal to stop any already running process. Already scheduled auto commits for e.g. in 2 minutes are stopped due to the `plugin.automaticsManager.unload();`

Gewerd-Strauss:
Sounds good to me, that's certainly a more elegant solution to the issue than my proposal :P

I don't see any issues with that from my end.

Vinzent03:
Thanks for your work on the initial implementation! Even though the final implementation is now quite different.

Gewerd-Strauss:
I don't care if it's not my code if my code is not optimal/if a more robust solution is found.
I mean, I set out saying that that probably wasn't the best way to do it, and now we have something better. 

Thank you  for the implementation.

PR #858, opened 2025-02-13
Fix diff view bug 

### LINKED ISSUE #857: [Bug]: Diff view fails and stops working on an edge-case
### Describe the bug

if you try to open a diff view of a file that's actually a nested repository (not a sub module), this plugin will throw an unhandled error, and then afterwards the diff view won't load any other files that you want to open.

A reload of the plugin is needed to reset this.

It's because only certain errors are handled in the gitShow function and others are just thrown and not being handled anywhere. (Is this intentional?)

```shell

```

### Steps to reproduce

1. Move a git repo inside your vault.
2. Commit it a few times.
3. Try to open it with the diff view.

### Expected Behavior

_No response_

### Addition context

I've done a bit of research on if there's a easy efficient git way to detect if any of the committed objects are actually a git repository, or other bad things like symlinks, but didn't find anything simple.

### Operating system

macOS

### Installation Method

None

### Plugin version

2.31.1

shumadrid:
Closes #857  Handles only the "nested repository" case, could be expanded to handle all errors.

### DIFF
--- a/src/ui/diff/splitDiffView.ts
+++ b/src/ui/diff/splitDiffView.ts
@@ -173,8 +173,13 @@ export default class SplitDiffView extends ItemView {
                 if (
                     error.message.includes("does not exist") ||
                     error.message.includes("unknown revision or path") ||
-                    error.message.includes("exists on disk, but not in")
+                    error.message.includes("exists on disk, but not in") ||
+                    error.message.includes("fatal: bad object")
                 ) {
+                    // Occurs when trying to run diff with an object that's actually a nested respository
+                    if (error.message.includes("fatal: bad object")) {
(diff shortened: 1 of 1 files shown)


shumadrid:
Update: I've noticed sadly that this plugin intercepts most of git errors and you are unable to handle them from your own plugin if you are interacting with git through this plugin.

Would you be open to changing this?
For example i want to call getRelativeRepoPath() and handle passing non-existent paths gracefully but currently it all results in a fatal error inside this plugin, and its breaks most functionality until you restart.

Just exposing a simple .raw() function without any error handling would also be good enough.

Vinzent03 (approved):

PR #864, opened 2025-02-19
Make view header buttons sticky and contain scrollbar to body

### LINKED ISSUE #863: View UI is inconsistent with Obsidian
The scrollbar stretches along the whole view instead of only the body and the header buttons aren't sticky when you scroll.

shumadrid:
Closes #863


Vinzent03:
Great idea, thanks for your contribution!

PR #884, opened 2025-03-21
fix: improve the "is file openable in obsidian" check

### LINKED ISSUE #883: [Bug]: "Is file openable by Obsidian" logic is incomplete
### Describe the bug

Currently the code for checking if a file can be opened in Obsidian isn't 100% correct.
e.g. if other plugins register extensions that they can open, this doesn't detect those.

There is a one-liner solution to make it more robust (maybe even 100% correct) but it involves reaching for a function that's inside the internal Obsidian API. 

```shell

```

### Steps to reproduce

.

### Expected Behavior

_No response_

### Addition context

_No response_

### Operating system

macOS

### Installation Method

None

### Plugin version

2.32.0

shumadrid:
Closes #883 

Gracefully handles the case where the internal API function isn't available anymore.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -321,10 +321,19 @@ declare module "obsidian" {
         saveLocalStorage(key: string, value: string | undefined): void;
         openWithDefaultApp(path: string): void;
         getTheme(): "obsidian" | "moonstone";
+        viewRegistry: ViewRegistry;
     }
     interface View {
         titleEl: HTMLElement;
         inlineTitleEl: HTMLElement;
+    }
+    interface ViewRegistry {
+        /**
+         * PRIVATE API
+         *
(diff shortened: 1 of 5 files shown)


shumadrid:
Stopped trying to bypass the lint check. ViewRegistry is a big interface and I don't want to add it to the obsidian types in `src/types.ts`

Vinzent03:
Yeah making eslint and typescript lsp happy is not always easy. I've now added the type to my private api collection...
Thanks for finding this way to check this better!

shumadrid:
No problem 😃, just trying to give back, since the existence of this plugin made developing my plugin a lot easier. 
That's a nice description of what the function does, i forgot to add one lol bc of all the fighting with eslint.

PR #886, opened 2025-03-24
fix: improve settings ui

### LINKED ISSUE #882: [FR] Some settings UI improvements
I did see a recent #878 issue that already made improvements. I have some more suggestions:

1. Show invalid textbox states for invalid settings instead of sending notices (which were deleted in #880 PR, but in the current implementation you can still specify negative numbers)
2. Set all default settings as placeholders (even after save, by naively checking if the saved setting matches the default one, and then setting a placeholder instead of .setValue())
3. Set a 80ms delay between saving the toggle settings that enable other settings, and refreshing the display, so that the toggle doesn't jump from one state to another but looks at least a bit like the others (imperfect fix but easy)

I can submit a PR for this if the maintainer agrees.

shumadrid:
Closes #882 

I didn't remove the default date formats from being applied as value because they're useful as a starting point.

Also fixed an inconsistency where the lowest possible Source control view refresh interval would be saved if the input was invalid (saves the default instead)

The setting "Show Stage and unstage Button in File Menu" was misleading because it also adds the "Add to gitignore" action, so i changed the setting to mention that.


Vinzent03:
I like still setting it for the date field. But what about the commit message text fields? I think it would be great if they wouldn't be set if equal to the default.

shumadrid:
It's up to you, i can change the date fields and the commit message text fields to be set as placeholders if you want. 
And I do agree with you for the commit messages.

Vinzent03:
I would leave the date fields as they are, but change the commit message text fields to the new style.

shumadrid:
had to change
`Omit<ObsidianGitSettings, "autoCommitMessage">`
to
`ObsidianGitSettings`
for the default settings btw

shumadrid:
`mergeSettingsByPriority()` still has that `Omit<ObsidianGitSettings, "autoCommitMessage">` type
didn't want to touch that

PR #890, opened 2025-03-28
fix: improve "Add to .gitignore" logic

### LINKED ISSUE #889: Suggested improvements to "Add to .gitignore" logic
- Add absolute paths instead of relative: since the function is used in the context of a certain file or folder context menu, i think it's logical to limit the scope of the added rule to only that item, and not accidentally ignore some other items.
- escape characters like `!` so they don't get treated as special characters.

shumadrid:
Closes #889


Vinzent03:
Good catch to fix this and taking care of many edge-cases! Thanks.

PR #893, opened 2025-04-01
feat: collapse unchanged lines in split diff view

### LINKED ISSUE #881: [Feature Request] Auto scroll to first diff in diff view
In long documents it is cumbersome having to scroll over a ton of unchanged text to get to (and find) the actual change in the diff. It would be nice to directly scroll to the first change in the diff. Even better would be a way to jump from one diff to another (next/previous).

shumadrid:
Closes #881 

And believe me, the original styling of the indicator was even uglier.

screenshots

### DIFF
--- a/src/ui/diff/splitDiffView.ts
+++ b/src/ui/diff/splitDiffView.ts
@@ -359,8 +359,8 @@ export default class SplitDiffView extends ItemView {
                 b: bState,
                 a: aState,
                 collapseUnchanged: {
-                    minSize: 5,
-                    margin: 3,
+                    minSize: 6,
+                    margin: 4,
                 },
                 diffConfig: {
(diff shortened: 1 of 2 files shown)


Vinzent03:
What do you think of this design? 
image

```css
/* Override styling of Codemirror merge view "collapsed lines" indicator */
.git-split-diff-view .ͼ2 .cm-collapsedLines {
    background: var(--interactive-normal);
    border-radius: var(--radius-m);
    color: var(--text-accent);
    font-size: var(--font-small);
    padding: var(--size-4-1) var(--size-4-1);
}
.git-split-diff-view .ͼ2 .cm-collapsedLines:hover {
    background: var(--interactive-hover);
}
```
It uses more standard css variables.

PR #922, opened 2025-06-15
setting panel convertion to obsidian-settings-ui

dariuszkanicki:
Hi,
I've been using Obsidian quite extensively lately and recently started developing my own plugin. That’s how I came across your plugin. In fact, it was one of the reasons I decided to develop this library — I found the settings panel a bit overloaded and, at times, confusing. I saw this as a great opportunity to improve the experience and create a reusable component that could benefit both myself and others in their own plugins.  
Of course, I appreciate your feedback and hope you’ll consider accepting the pull request.


Vinzent03:
For now, I would prefer to keep this as is and not depend on an external dependency. The requirements might change fast and don't want to depend on another dependency for that. If more other plugins start using this and is more established I might switch though.

dariuszkanicki:
I understand your reasoning, but to be honest, I'm quite disappointed. Have you taken the time to actually download, build, and run the converted version? I put serious effort into fully adapting your plugin—not just to prove a point, but to demonstrate the practical value from both a developer and user perspective. The improvements are tangible, especially when you experience the settings UI firsthand.

That’s why I find your explanation a bit hard to follow. You mentioned that “requirements might change fast,” which makes you hesitant to rely on a library—but I’m not sure which specific requirements you’re referring to, or how they would actually be limited by using this shared abstraction. The library is designed to be flexible and extensible. If anything, centralizing UI logic in one place makes it easier to adapt to future needs—not harder.

You also said you might adopt it once “more plugins use it and it’s more established.” But that creates a catch-22: if everyone waits for others, the library will never gain traction. Someone has to lead by example, and your plugin is a textbook use case where this library clearly improves clarity, structure, and maintainability. Right now, the settings are so dense that it's hard to see the forest for the trees. The refactored version solves that very directly.

To support this further, I’ve also written an article detailing the entire process—why I did the conversion, the design decisions behind it, and what problems it solves. I’ll be publishing it on LinkedIn shortly. Your current stance will naturally be part of the narrative, and I genuinely hope it turns into a positive example of embracing a thoughtful and collaborative solution—rather than a missed opportunity.

**Given all this, I would really encourage you to take a second look and reconsider your position—not at some future point, but now, based on what’s already in front of you. The conversion speaks for itself, and this could be a chance to set a positive precedent rather than waiting for one.**

PR #941, opened 2025-07-14
fix: Prevent jitter of refresh icon in mobile

### LINKED ISSUE #916: [Bug]: Refresh Icon Jitter Issue in Git Source Control Toolbar
### Describe the bug

**Description**:

When using the Git plugin in Obsidian Mobile (Android), the refresh icon (🔄) in the bottom toolbar exhibits noticeable jitter/shaking upon clicking.

**Steps to Reproduce**:

1. Open Obsidian Mobile on an Android device
2. Navigate to a vault with Git plugin enabled
3. Click the refresh icon in Git Source Control's bottom toolbar
4. Observe the icon jitter phenomenon

**Additional Information**:
- Issue occurs only on mobile (desktop version works normally)
- Screen recording: https://github.com/user-attachments/assets/14a09ceb-4fb9-469e-ae01-30a1574e24b5

```shell

```

### Steps to reproduce

1. Open Obsidian Mobile on an Android device
2. Navigate to a vault with Git plugin enabled
3. Click the refresh icon in Git Source Control's bottom toolbar
4. Observe the icon jitter phenomenon

### Expected Behavior

The position of the icon remains unchanged.

### Addition context

The icon jiggles from side to side. 

### Operating system

Android

### Installation Method

None

### Plugin version

2.33.0

SHNakajima:
Closes #916

The root cause was that the .nav-buttons-container did not have a fixed minimum height, causing layout shifts when the loading animation started

Tested on:
- Obsidian Mobile (Android Pixel 8): Jitter is resolved.
- Obsidian Mobile (Android Emulator on Bluestack 720 x 1080 pixel): Jitter is resolved.
- Obsidian Desktop: No regression observed.

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -17,12 +17,12 @@
     padding: 0;
 }
 
-.workspace-leaf-content[data-type="git-view"] .nav-buttons-container {
-    min-height: 34px;
-}
-
 .workspace-leaf-content[data-type="git-history-view"] .view-content {
     padding: 0;
+}
+
+.loading {
+    overflow: hidden;
 }
 
 .loading > svg {


Vinzent03:
I think using the variable `--icon-size` should be better. Could you test whether the that fixes the issue as well?

https://docs.obsidian.md/Reference/CSS+variables/Foundations/Icons

SHNakajima:
Thank you for checking on this!

I tried using the `--icon-size` variable as you suggested, but unfortunately it didn't resolve the issue.

However, I discovered that simply adding `.loading { overflow: hidden; }` to the original `styles.css` (before this PR's changes) actually fixes the problem completely. This seems like a much cleaner solution.

Let me know what you think about this approach!

Vinzent03:
Okay great. Thanks for the fix!

PR #969, opened 2025-09-23
Bug Fix: (Tiny) Text Area Quotation 

Andrew-McCall:
In the settings, when the saved commitMessage or autoCommitMessage (Text Areas) load a saved value, it adds quotes to the start and end. If you edit the message and don't delete the quotes, it "correctly" escapes the quotes in the field and saves and repeats.

(See image at end, if I'm not being clear)

When loading the value, I've changed it to use String instead of JSON.Stringify so strings are passed as normal, only downside would be an object would show as [object] but this text value would never hold an object, just strings.

This is a tiny fix, No AI. umm. no unit tests either.

Thank you for your time (and great plugin).

image

### DIFF
--- a/src/setting/settings.ts
+++ b/src/setting/settings.ts
@@ -1316,7 +1316,7 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
 
         if (defaultValue !== storedValue) {
             // Doesn't add "" to saved strings
-            if (typeof storedValue === "string") {
+            if (typeof storedValue === "string" || typeof storedValue === "number" || typeof storedValue === "boolean") {
                 text.setValue(String(storedValue));
             }else {
                 text.setValue(JSON.stringify(storedValue));


Andrew-McCall:
In case I'm just rambling, when the saved commit message value is loaded, it would have erroneous quotes added to it in the text boxes.

Andrew-McCall:
image

Andrew-McCall:
My first commit only used String() inplace of JSON.stringify as I wouldn't expect the stored value to ever be an object however the linter wants a more defensive stringify. 

Made my intent more clear, if it's a string (or a value that displays as a string - number, bool, string), we String() it. Else its an unexpected value - object, null or undefined, it uses JSON.stringify which matches the other setter's behaviour.

Vinzent03:
Thanks for the fix! I don't actually think there are values, which are not just string or number in the settings, but this should be solid now.
I somehow messed up the merge, but did it locally now. 806da55d32fe35455782a86ad86493070c727e95

PR #995, opened 2025-12-03
Update Authentication.md

patchuby:
Removed extraneous "set" from git config command


Vinzent03:
The current version is preferred as the one without `set` is deprecated: https://git-scm.com/docs/git-config#_deprecated_modes

patchuby:
Ah noted @Vinzent03. Then could you add a note that this might not work with previous versions of git ?
Latest git version on stable ubuntu is 2.34 (and other distros seem to follow like Pop-OS which I'm using). 
Using `git config set` does not work and yields a error message that's not useful in debugging the issue when looking on the web.

Vinzent03:
Oh okay. Thanks for letting me know then. Didn't know the `set` subcommand is so "new" and some Ubuntu lts ships such old git version.

PR #997, opened 2025-12-05
Add setting to show/hide git refs in history view

mike-kob:
When using git primarily for version control of notes in the vault rather than backup, it's not so useful to track where the refs (especially remotes) are, but they add clutter to the history view.

This MR adds a new setting "Show git refs" under "History view" heading defaulting to `true` (non-breaking change):
image

Behaviour with toggle on:
image

and with toggle off:
image

Note: as for other settings in "History view", changes take effect only after window reload.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -43,6 +43,7 @@ export const DEFAULT_SETTINGS: ObsidianGitSettings = {
     showFileMenu: true,
     authorInHistoryView: "hide",
     dateInHistoryView: false,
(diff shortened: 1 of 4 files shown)


Vinzent03:
I don't think it takes that much space, but is really helpful to understand your current repo state. I would suggest to use a custom css snippet like the following to hide this:
(Custom css snippets can be added and enabled in the Obsidian appearance settings)
```css
.git-ref {
    display: none;
}
```

Issue #1000, opened 2025-12-20
[Bug]: username prompt uses same modal as password prompt


UncleThaodan:
### Describe the bug

Both prompts are using `GeneralModal()` with no further configuration (so no change in `config.obscure` either), making inputting the username both confusing and annoying, as it is barely distinguishable from the password prompt.

### Relevant errors (if available) from notifications or console (`CTRL+SHIFT+I`)

```shell

```

### Steps to reproduce

Make the plugin prompt you for username/password.

### Operating system

Linux

### Installation Method

Other

### Plugin version

2.35.2

Vinzent03:
It does recognize the username request for English systems. I don't know of a solution to make this work language independent.

PR #1007, opened 2026-01-08
fix: obscure password prompt in isomorphic-git auth

AnrokX:
Adds obscure: true to the password prompt in isomorphicGit.ts, masking passwords/PATs during authentication.

  Missed in 5dbd3cd which fixed this for simpleGit.ts but not isomorphicGit.


Vinzent03:
Thanks for the fix, but this does not close #1000 as it is about username and password prompt being both obscure on desktop.

PR #1022, opened 2026-01-26
feat: allow empty default manual commit message

JKluckow:
When enabled, the commit message field stays empty and the user must provide their own message. Commits without a message are rejected with a notice.

### DIFF
--- a/src/setting/settings.ts
+++ b/src/setting/settings.ts
@@ -282,32 +282,30 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
 
             new Setting(containerEl).setName("Commit message").setHeading();
 
-            new Setting(containerEl)
+            const manualCommitMessageSetting = new Setting(containerEl)
                 .setName("Commit message on manual commit")
                 .setDesc(
                     "Available placeholders: {{date}}" +
                         " (see below), {{hostname}} (see below), {{numFiles}} (number of changed files in the commit) and {{files}} (changed files in commit message). Leave empty to require manual input on each commit."
-                )
-                .addTextArea((text) => {
-                    text.setValue(plugin.settings.commitMessage);
-                    text.onChange(async (value) => {
-                        plugin.settings.commitMessage = value;
-                        await plugin.saveSettings();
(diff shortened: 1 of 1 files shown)


Vinzent03:
What about the existing option to require a custom commit message on automatic commit and the dedicated commands to commit with specific commit message? I think this use case is already covered enough and does not need another setting.

benediktpfautsch:
The idea behind this feature is for people who don’t want to use Git purely as a backup system for their notes, but rather in the traditional Git workflow. When collaborating with several people and making changes, you’re expected to commit your work with a personal message explaining what you did.

The auto-commit feature, however, creates commits at fixed time intervals, regardless of whether you’ve actually finished working on a page or section.

Vinzent03:
I totally get that, but everything you need is already implemented. You can either disable automatics commit or enable the setting to get asked for a commit message on each automatic commit. For manual commit there is the additional command "Commit with specific message"  to pass a custom message before commit.

benediktpfautsch:
Thanks, I found the "Commit with specific message" commands in the command palette and they work as intended. But they only work via the command palette - not via the Source Control Panel button. In this panel, the message field is always pre-filled with the default message (e.g. vault backup: {{date}}) and resets to this default after every commit. There's no way to leave it empty or require a custom message there. This PR would solve exactly that. My Workaround atm is that I prefill "Commit message on manual commit" with a single space character, because git doesn't allow just a space character. But that's ugly.

Vinzent03:
Okay I get hat scenario. To avoid adding to many additional options, I think its cleaner to allow the existing manual commit message setting to stay empty and provide a button to reset it to the default value. Instead of always using the default when the setting is empty.

Vinzent03:
Thanks for the changes. I really like the new way it behaves and makes more sense now. I just changed it to use an icon instead of a text on the button and it now edits the textarea content instead of the setting and reloading the settings pane. If you are fine with changes, this is ready to go.

JKluckow:
Ok, that works for me. Thanks!

PR #1030, opened 2026-02-08
Added MacOS installation details

YuriySamorodov:
Expanded MacOS section in installation.md file and explained how Obsidian can be installed on Mac.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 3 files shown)


Vinzent03:
Installing git generally or especially a specific package manager, should not be the scope of this plugin. Git already lists all installation methods and links to homebrew as well. I think it should be better to just link to the official documentation: https://git-scm.com/install/mac

YuriySamorodov:
Hi, @Vinzent03 . Good point! Should We mention keychain-related details the same way it was done in Windows section?

Vinzent03:
Yeah I think those are fine and an improvement.

Issue #1032, opened 2026-02-09
[Bug]: Images are not rendering


YuriySamorodov:
### Describe the bug

Looks like images in instruction files are not rendering at all.
The following asstets are affected

- [Installation.md](docs/Installation.md)

### Relevant errors (if available) from notifications or console (`CTRL+SHIFT+I`)

```shell

```

#### Installation.md 

<img alt="Image" src="https://github.com/user-attachments/assets/53f56ebf-1899-4ebf-832b-707fc80046aa" width="55%" height="55%" />

I have noticed all image links in docs/Installation.md are in the following format:
```
![[assets/third-party-windows-git.png]]
```

instead of 
```
![](assets/third-party-windows-git.png)
```
as in docs/Line Authoring.md.

### Operating system

macOS

### Installation Method

None

### Plugin version

2.36.1

PR #1033, opened 2026-02-09
Fix image links (#1032)

YuriySamorodov:
Updated image links to  HTML tags to achive the following goals:

- Fix image rendering in Installation.md
- Handle larger images size to fit in screen without stretching

This is a solution for #1032

### DIFF
--- a/docs/Authentication.md
+++ b/docs/Authentication.md
@@ -16,7 +16,7 @@ You have to do one authentication action (clone/pull/push) after setting the hel
 
 To remove or update stored credentials, open Keychain Access, search for entries for your remote host (e.g. github.com), and delete or update the matching entry.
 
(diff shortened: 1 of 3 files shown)


Vinzent03:
The documentation is meant to work for Obsidian publish: https://publish.obsidian.md/git-doc/Installation
The images are rendered correctly and I don't think they need to be resized as the page is not full width itself.

YuriySamorodov:
> The documentation is meant to work for Obsidian publish: https://publish.obsidian.md/git-doc/Installation The images are rendered correctly and I don't think they need to be resized as the page is not full width itself.

Thanks for the explanation, @Vinzent03 ! Indeeed, the current format works in Obsidian, but the suggested one does not.

PR #1044, opened 2026-03-03
feat: Default {{hostname}} placeholder to OS hostname when not explicitly configured

### LINKED ISSUE #1043: {{hostname}} should default to OS hostname if not specified
The commit message supports {{hostname}} placeholder which can be explicitly set in the settings (and appears to be persisted somewhere besides the plugin's data.json). However, if {{hostname}} placeholder replacement is not set, the {{hostname}} placeholder just returns blank. It would be more helpful for this to default to the hostname from the OS so I don't have to set it unless I want to override to some other value.

aaron-meyers:
* feat: fall back to OS hostname when {{hostname}} placeholder is not configured

When the {{hostname}} placeholder is used in commit messages but no custom hostname is set in localStorage, fall back to the OS hostname via os.hostname(). This handles the case gracefully on mobile where the os module is unavailable.

Tested locally. Resolves #1043.

### DIFF
--- a/src/gitManager/gitManager.ts
+++ b/src/gitManager/gitManager.ts
@@ -1,3 +1,4 @@
+import { hostname as osHostname } from "os";
 import { type App, moment } from "obsidian";
 import type ObsidianGit from "../main";
 import type {
@@ -274,7 +275,14 @@ export abstract class GitManager {
             template = template.replace("{{numFiles}}", String(numFiles));
         }
         if (template.includes("{{hostname}}")) {
-            const hostname = this.plugin.localStorage.getHostname() || "";
+            let hostname = this.plugin.localStorage.getHostname() || "";
+            if (!hostname) {
+                try {
+                    hostname = osHostname();
+                } catch {
+                    // os module not available (e.g., on mobile)
+                }
+            }
             template = template.replace("{{hostname}}", hostname);
         }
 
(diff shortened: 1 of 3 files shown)


src/setting/settings.ts
```diff
@@ -346,7 +346,9 @@ export class ObsidianGitSettingsTab extends PluginSettingTab {
 
             new Setting(containerEl)
                 .setName("{{hostname}} placeholder replacement")
-                .setDesc("Specify custom hostname for every device.")
+                .setDesc(
+                    "Specify custom hostname for every device. Defaults to the OS hostname if not set."
```
Vinzent03:
```suggestion
                    "Specify custom hostname for every device. Defaults to the OS hostname if not set on desktop."
```

Vinzent03 (changes requested):
Thanks for the pr! I just left a small remark.

src/gitManager/gitManager.ts
```diff
@@ -274,7 +275,14 @@ export abstract class GitManager {
             template = template.replace("{{numFiles}}", String(numFiles));
         }
         if (template.includes("{{hostname}}")) {
-            const hostname = this.plugin.localStorage.getHostname() || "";
+            let hostname = this.plugin.localStorage.getHostname() || "";
+            if (!hostname) {
+                try {
```
Vinzent03: A better way would probably be to use `Platform.isDesktopApp` instead of the try catch

aaron-meyers: Yeah I'll do that. I'm not actually sure if os.hostname() is available on mobile or not (Copilot doesn't seem to think it would work but I haven't tried it) but given that obsidian-git is not primarily targeted for use on mobile it seems fine to just scope this to desktop.

Vinzent03 (approved):
LGTM!

PR #1052, opened 2026-03-16
fix: prevent  "detached HEAD" state after using "Switch to remote branch" command

reinaldoarrosi:
When using the "Switch to remote branch" command the git repository goes into "detached head" mode because obsidian-git is calling `git checkout origin/<branch_name>` which does not create a local branch that tracks the remote branch.

The fix for it is simple: if a local branch tracking the remote branch does not exists yet, use `git checkout -b <branch> origin/<branch>` to checkout the remote branch AND create a local branch with the same name that tracks the remote branch.

If a local branch with the same name already exists AND it tracks the remote branch, then simply call `git checkout <branch>`

### DIFF
--- a/src/gitManager/simpleGit.ts
+++ b/src/gitManager/simpleGit.ts
@@ -901,6 +901,8 @@ export class SimpleGit extends GitManager {
     }
 
     private async getLocalBranchUpstream(localBranchName: string): Promise<string | undefined> {
+        // Returns the configured upstream ref for a local branch (e.g. `origin/main`), using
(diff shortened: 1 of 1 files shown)


Vinzent03:
Looks great to me. Interesting idea to just iterate over multiple branch names to always get a fresh branch for that remote. I think that is fine. 
I would just prefer a bit more doc comments for the methods on what they do. Especially `getAvailableLocalBranchName`.

reinaldoarrosi:
I just pushed a commit that adds comments to better explain the whole branch checkout flow

PR #1063, opened 2026-04-06
feat:add chinese support

CY1211NET:
I am a student from China, and I strongly support this plugin, so I added Chinese support to it, hoping to gain recognition.


Vinzent03:
Just opening a pull request before discussing such a huge change in an issue and don't providing any description of how this is implemented is not helpful. I'm unsure yet how and if to support translations, but that would be discussed in an issue first.

PR #1064, opened 2026-04-08
fix: set GIT_WORK_TREE when using custom git directory

dragonfax:
I have a scenario where I don't want the .git directory to be inside the vault. The reason is because I also have Google Drive syncing this vault Automatically. And I don't want Google Drive to sync the Git files. (Google Drive doesn't have any sort of files-to-ignore configuration)

Unfortunately everything I tried wasn't working. I know this plugin already has support for different Git repo, but with an external git repo you have to specify both the GIT_DIR and the GIT_WORK_TREE explicitly in the environment or through command line options. And this plugin doesn't support that. 

This PR fixes that by adding support for the GIT_WORK_TREE. It's not necessary to specify it explicitly as it will default to the vault/base directory. It's harmless to have it present in other scenarios because it will still have the correct value, the base path. 

> [!NOTE]
> Running `git` from within the working directory (basepath) used to be enough, some 15 years ago, but no longer. Also this is completely unrelated to `git worktree` support.

### DIFF
--- a/src/gitManager/simpleGit.ts
+++ b/src/gitManager/simpleGit.ts
@@ -82,6 +82,7 @@ export class SimpleGit extends GitManager {
             }
             if (gitDir) {
                 envs["GIT_DIR"] = gitDir;
+                envs["GIT_WORK_TREE"] = basePath;
             }
             for (const envVar of envVars) {
                 const [key, value] = envVar.split("=");


dragonfax:
cc @Vinzent03 @reinaldoarrosi

Vinzent03:
What was the exact issue you were experiencing? Reading the documentation it should be enough to launch the git process from the vault base path which is the case currently: 

> Set the path to the repository (".git" directory). This can also be controlled by setting the GIT_DIR environment variable. It can be an absolute path or relative path to current working directory.

> Specifying the location of the ".git" directory using this option (or GIT_DIR environment variable) turns off the repository discovery that tries to find a directory with ".git" subdirectory (which is how the repository and the top-level of the working tree are discovered), and tells Git that you are at the top level of the working tree. If you are not at the top-level directory of the working tree, you should tell Git where the top-level of the working tree is, with the --work-tree=<path> option (or GIT_WORK_TREE environment variable)

> If you just want to run git as if it was started in <path> then use git -C <path>.

dragonfax:
The issue I was having was that I have my Git repository outside of the obsidian vault path, and no configuration options would get the plugin to work with that scenario.

dragonfax:
My vault is "/Users/jstillwell/Documents/Obsidian Vault" 
my bare git repo is "/Users/jstillwell/Documents/obsidian.git"

I have the custom git path set.
Screenshot 2026-04-08 at 5 06 17 PM

But that alone didn't work.

And no settings for "Additional Environment Variables" helped either. GIT_DIR and/or GIT_WORK_TREE.

I'd always get 
Screenshot 2026-04-07 at 11 07 30 PM

But applying this one line patch fixes everything. Now I don't even need the environment variables either. Just the custom git path.

Does that explain it?

Vinzent03:
Hmm it works perfectly fine for me without that setting. Which git version are you running? (`git --version`)

dragonfax:
```
jstillwell@MacBook-Pro ~ % git --version
git version 2.50.1 (Apple Git-155)
```

Vinzent03:
I could finally reproduce the case. So far, I only tested it by moving the `.git` directory, but that is not a real bare git repo. So by making it bare, the work tree env var is needed. Thanks for fixing it!

dragonfax:
Yeah, it's a weird pattern, I know. Using a bare repo lets you keep your git work tree completely separate from the git files. Useful if you have some tooling that works with the directory, and you don't want that fooling with the repo/git history. Or if you’re serving those files up, say, via a web server, and you don't want people to access the history either.

Thanks for the merge.

Also, I didn't know that a `.git` dir wasn't also a bare repo. That's new. I looked into it, and yeah, the repos save whether they were created as `--bare` or not. Interesting.

PR #1068, opened 2026-04-18
[Enhancement]: added css variables for styling

theinfinull:
exposed overridable CSS custom properties under
.theme-dark and .theme-light for user-styling
via snippets

includes variables for:
- file-status markers for modified and deleted
- diff view line backgrounds/highlights
- editor gutter markers

Docs updated with variable reference

### DIFF
--- a/docs/Start here.md
+++ b/docs/Start here.md
@@ -15,6 +15,7 @@ aliases:
 -   [[Tips-and-Tricks|Tips-and-Tricks]]
 -   [[Common issues|Common Issues]]
 -   [[Line Authoring|Line Authoring]]
+-   [[Styling|Styling]]
 
 > [!warning] Obsidian installation on Linux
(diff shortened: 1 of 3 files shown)


Vinzent03:
I'm unsure if adding variables is really necessary/the right solution. I would suggest you to write css rules with a higher [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity), which should be possible with for example `!important`

theinfinull:
`!important` works, but it forces users to replicate internal selectors. like overriding the gutter marker size currently needs **~8** `!important` rules. All of which may silently break if those class names ever change or if new ones are added

Variables collapse em into one line config, and users can style via docs without knowing internal class names, which is [how obsidian itself does it](https://docs.obsidian.md/Reference/CSS+variables/CSS+variables).

Seemed like a cleaner approach - open to feedback

UbeJelly:
This is nice, it's easy to handle for both new/casual and experienced Obsidian users

PR #1081, opened 2026-05-07
fix: improve mobile debug and credential handling

garnetlyx:
## Summary
- add an opt-in debug logging setting that writes redacted diagnostics to `plugin/log/git-log.md`
- mask the stored password/token field in settings and fix migration of the legacy username setting
- make the mobile isomorphic-git adapter throw proper `ENOENT` errors for missing reads/stat calls instead of returning null-like values or bare objects

## Duplicate check
- checked open PRs and recent merged PRs; no open PR covers debug logging, settings password/token masking, or mobile adapter missing-read `ENOENT` handling
- related merged PR #1007 only obscures the isomorphic-git auth prompt, not the stored settings password/token field

## Test plan
- `pnpm run tsc`
- `pnpm run svelte`
- `pnpm run format`
- `pnpm run lint`
- `pnpm run build`

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -22,6 +22,7 @@ export const DEFAULT_SETTINGS: ObsidianGitSettings = {
     pullBeforePush: true,
     disablePopups: false,
     showErrorNotices: true,
+    debugLogging: false,
     disablePopupsForNoChanges: false,
     listChangedFilesInMessageBody: false,
     showStatusBar: true,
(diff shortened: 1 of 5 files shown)


Vinzent03:
What is the use case for this specific logging? I would not work on the github token setting at the moment since I would migrate to the new Obsidian secret api to store the token app wide and more securely.

PR #1096, opened 2026-06-12
Feat/i18n chinese support

zhou-yusen:
This PR adds internationalization (i18n) support to the plugin, with Simplified Chinese as the first non-English language.

## Changes
- Added lightweight i18n system using Obsidian's `getLanguage()` API (no new dependencies)
- Extracted ~400+ user-facing strings into locale files (`src/locale/`)
- Added English (`en.ts`) and Simplified Chinese (`zh.ts`) translations
- Added language selector dropdown in Miscellaneous settings (auto/en/zh)
- Language preference is persisted in plugin settings

## Design Decisions
- Used a custom lightweight approach (same pattern as obsidian-kanban) instead of i18next to minimize bundle size
- Git terminology (commit, push, pull, etc.) is kept in English in the Chinese translation, as is conventional in the Chinese developer community
- English remains the fallback for any missing translations
- All existing functionality is preserved; no behavioral changes

## Files Changed
- 3 new files: `src/locale/index.ts`, `src/locale/en.ts`, `src/locale/zh.ts`
- 19 modified files with string replacements


singlezz:
这个好，支持

remnant-song:
支持支持

PR #1090, opened 2026-05-23
fix: binary writes in mobile git adapter

TadokoroYuki:
## Summary

This normalizes binary data before passing it to Obsidian's binary file APIs in the isomorphic-git adapter.

`isomorphic-git` can call the fs `writeFile` implementation with `Buffer` / `Uint8Array` views for pack/index data, while Obsidian's `writeBinary` and `modifyBinary` APIs expect an `ArrayBuffer`. The adapter now converts `ArrayBufferView` values to an exact `ArrayBuffer` before writing them.

## Why

Recent `isomorphic-git` versions added packfile integrity validation. That can surface previously silent binary persistence problems as errors like:

`Packfile payload corrupted: calculated ... but expected ... The packfile may have been tampered with.`

This is relevant to mobile clone/fetch failures reported in #1085 and related mobile clone/fetch reports in #642.

## Validation

- `pnpm run tsc`
- `pnpm run svelte`
- `pnpm run lint`
- `pnpm run format`
- `pnpm run build`
- Ran a local adapter-boundary check that bundles `MyAdapter` and verifies `Uint8Array` / `Buffer.subarray()` writes are converted to exact-length `ArrayBuffer` values.
- Ran a local smart-HTTP fixture with `isomorphic-git@1.37.4` through `MyAdapter`: shallow clone and fetch completed, with `.pack` and `.idx` written via an Obsidian-like `writeBinary` implementation that rejects non-`ArrayBuffer` values.

## Mobile testing

This has now been verified on a real mobile Obsidian environment by installing the fork test build and retrying the clone/fetch flow that previously failed with `Packfile payload corrupted`.

To make mobile verification easier, I published a fork pre-release with `main.js`, `manifest.json`, and `styles.css`:

https://github.com/TadokoroYuki/obsidian-git/releases/tag/codex-mobile-packfile-test-a0509f1

Additional device coverage is still welcome, especially across both iOS and Android clone/fetch paths.

PR #1098, opened 2026-06-17
fix: prevent corrupted packfiles from compressed fetch responses on mobile

Narcolepsyy:
## Problem

On mobile, fetch/clone/pull can fail with:

```
An internal error caused this command to fail.
... Packfile payload corrupted: calculated <sha> but expected <sha>.
The packfile may have been tampered with.
```

Packfile payload corrupted error on mobile

*Error shown on mobile after a fetch, before this fix. Raised by isomorphic-git at the packfile payload-SHA verification step.*

This is thrown by isomorphic-git's packfile integrity check — but the corruption originates in the **HTTP receive path**, not the filesystem adapter (so it is unrelated to, and not covered by, the binary-write normalization in #1090).

The smart-HTTP response is collected through Obsidian's `requestUrl` and handed to the packfile parser as an `ArrayBuffer`. If a server or proxy returns a **gzip-compressed** body and the platform does not transparently inflate it, the still-compressed (or otherwise altered) bytes get written as a packfile that passes the trailer check but fails the payload SHA later, when its objects are read. A valid git smart-HTTP body begins with an ASCII pkt-line length or `PACK` — never the gzip magic bytes — so a body starting with `1f 8b` is unambiguously a response that should have been inflated.

## Fix

Harden the HTTP client in `getRepo()` in two layers:

1. **Request identity encoding** — send `Accept-Encoding: identity` so the server is asked not to compress the response. Caller-supplied headers still override it.
2. **Detect-and-inflate (defense in depth)** — if the response body still starts with the gzip magic bytes (`0x1f 0x8b`), inflate it with `DecompressionStream("gzip")` before handing it on.

The inflate is keyed on the **actual body bytes, not the `Content-Encoding` header**, on purpose: some platforms transparently inflate the body while leaving the header in place, and trusting the header would double-inflate and *introduce* the very corruption this fixes. If decompression is unavailable or fails, it falls back to the original bytes, so behavior is never worse than before.

## Testing

- `pnpm run all` (tsc + svelte-check + prettier + eslint) and `pnpm run build` all pass.
- The `inflateIfGzipped` logic was exercised in a Node harness:
  - gzipped packfile-like payload → inflated back to the exact original bytes;
  - plain pkt-line body and bodies starting with `PACK` → passed through unchanged;
  - empty and single-byte buffers → no crash, passed through.

## Notes / caveats

- This prevents *new* corruption on the receive path. It does not repair a pack that is already corrupt on disk or on the remote — for that, a re-clone (or server-side repack) is still required. In the reporter's case a re-clone restored normal operation, which is consistent with a transient compressed/truncated transfer.
- I can't exercise the full Obsidian mobile networking stack from here, so on-device verification of clone/fetch against the affected remote would be a good final check before merge.
- Separate from #1097 (the criss-cross merge fallback); these touch different code paths and can be reviewed independently.

### DIFF
--- a/src/gitManager/isomorphicGit.ts
+++ b/src/gitManager/isomorphicGit.ts
@@ -119,16 +119,33 @@ export class IsomorphicGit extends GitManager {
                     const res = await requestUrl({
                         url,
                         method,
-                        headers,
+                        // Ask the server not to compress the response. The
+                        // packfile parser needs the exact raw bytes; if a
+                        // server or proxy gzips the response and the platform
+                        // does not transparently inflate it, the packfile is
+                        // persisted corrupted and only fails later with
+                        // "Packfile payload corrupted".
+                        headers: { "Accept-Encoding": "identity", ...headers },
                         body: collectedBody,
                         throw: false,
                     });
+
+                    // Defense in depth: if the response still arrived gzipped
+                    // (a server/proxy ignored the request and the platform did
+                    // not inflate it), inflate it ourselves. This is keyed on
+                    // the gzip magic bytes rather than the `Content-Encoding`
+                    // header on purpose: some platforms auto-inflate but leave
+                    // the header in place, and trusting it would double-inflate
+                    // and corrupt the body.
+                    const responseBuffer = await inflateIfGzipped(
+                        res.arrayBuffer
+                    );
 
                     return {
                         url,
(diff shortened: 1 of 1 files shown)


Vinzent03:
Thanks for the contribution!

PR #1103, opened 2026-06-27
fix: avoid misleading "Committed 0 files" notice

obook:
When committing often, commit-and-sync sometimes showed "Committed 0 files".

simple-git resolves commit() with { changes: 0 } instead of throwing when there is nothing left to commit, so a no-op commit was reported as a real one. This now shows "No changes to commit" in that case, which also respects the "hide notifications for no changes" setting.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1060,11 +1060,19 @@ export default class ObsidianGit extends Plugin {
                     committedFiles =
                         unstagedFiles.length + stagedFiles.length || 0;
                 }
-                this.displayMessage(
-                    `Committed${roughly ? " approx." : ""} ${committedFiles} ${
-                        committedFiles == 1 ? "file" : "files"
-                    }`
-                );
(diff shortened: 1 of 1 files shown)


Vinzent03:
Did you experience this on a regular basis. If so in what scenario? By automatic commits or manual ones? The method should prevent committing if there are no changes by the result of git status.

obook:
I have seen it on Linux a few times but could not reproduce it reliably, so let us set this PR aside for now. I will revisit it with a reproduction if it comes up again. That said, the change is harmless on its own ("No changes to commit" is accurate whenever changes === 0, reproducible or not), so feel free to keep it around rather than close it. 

Either way works, thanks!

Vinzent03 (approved):
Yeah you are right. Adding these changes might be nevertheless an easy fix. Thanks for your contribution!

PR #1105, opened 2026-06-27
feat: add option to commit-and-sync before quitting

obook:
Adds a "Commit-and-sync before quitting Obsidian" toggle.

When enabled, the plugin runs a commit-and-sync on the quit event, so the last local changes are pushed before Obsidian closes. It uses the official quit hook and has a timeout so a hanging network does not block quitting. This is best effort, since the quit event is not guaranteed to run on a crash or a forced quit.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -20,6 +20,7 @@ export const DEFAULT_SETTINGS: ObsidianGitSettings = {
     autoCommitOnlyStaged: false,
     disablePush: false,
     pullBeforePush: true,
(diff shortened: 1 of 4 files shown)


Vinzent03:
As discussed in this [issue](https://github.com/Vinzent03/obsidian-git/issues/13), this method has many issues related to timeouts provided by the os that can easily lead to a corrupted state. So I very much favor using the provided command to commit and sync and then close Obsidian.

obook:
Thanks for the detailed explanation.

PR #1112, opened 2026-07-10
perf: optimize getTreeStructure to run in linear time

mrueg:
### Cause of the Issue
  The issue lay in the  getTreeStructure  function within gitManager.ts.
  Whenever the file list in the source control or history views is updated, the plugin constructs a directory tree structure of the changed files. The original implementation  _getTreeStructure  did this recursively with a loop that:

  1. Filtered all items matching a path prefix in each iteration.
  2. Removed items from the array using  .remove() , which is an O(N) operation inside a loop.
  3. Recursively walked directories.

  This resulted in an overall worst-case complexity of O(N²) to O(N³), which dramatically degraded performance as the size of the changeset grew, eventually freezing/hanging the main Obsidian thread.

  ### Solution

  I replaced the recursive implementation with a Trie-based tree builder that constructs the directory tree structure in linear time O(N·L) (where N is the number of files and L is the path depth, which is typically small):

  • Paths are split and mapped to nested dictionaries (Tries) in a single linear pass.
  • The Trie is converted back into a list structure in a single linear pass.
  • The original simplification and sorting algorithms ( simplify ) are applied to maintain identical output structure and ordering.

  ### Benchmark Results

  Running a comparison test with 20,000 files in distinct directories:

  • Original recursive approach:  7.885 seconds  (causing the UI to completely freeze and throw unresponsive prompts).
  • Optimized Trie-based approach:  43.546 milliseconds  (virtually instantaneous and runs smoothly within a single frame).

### DIFF
--- a/src/gitManager/gitManager.ts
+++ b/src/gitManager/gitManager.ts
@@ -162,51 +162,6 @@ export abstract class GitManager {
 
     unload(): void {}
 
-    private _getTreeStructure<T = DiffFile | FileStatusResult>(
-        children: (T & { path: string })[],
-        beginLength = 0
-    ): TreeItem<T>[] {
-        const list: TreeItem<T>[] = [];
-        children = [...children];
-        while (children.length > 0) {
-            const first = children.first()!;
-            const restPath = first.path.substring(beginLength);
-            if (restPath.contains("/")) {
-                const title = restPath.substring(0, restPath.indexOf("/"));
-                const childrenWithSameTitle = children.filter((item) => {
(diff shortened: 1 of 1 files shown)


Vinzent03 (changes requested):
Thanks for your contribution and the great improvement. Please remove the added sass package and this looks good to me.

mrueg:
@Vinzent03 thanks, removed sass there.

Vinzent03 (approved):

PR #1113, opened 2026-07-10
perf: debounce line authoring trackChanged on file modification

mrueg:
### Performance Improvement: Debouncing Line Authoring modify events

  • Bottleneck: Inside lineAuthorIntegration.ts, the plugin was listening to the vault  "modify"  event to update line authoring info. This event fires on every single keypress as you type. Without debouncing, the plugin was spawning Git sub-processes (such as  git rev-parse HEAD  and  git hash-object ) on every keypress, causing substantial CPU usage and input latency during
editing.
  • Fix: Added a 1000ms debouncer to the  "modify"  event handler. Git processes are now only spawned after the user pauses typing for 1 second, dramatically reducing CPU overhead while editing.

### DIFF
--- a/src/editor/lineAuthor/lineAuthorIntegration.ts
+++ b/src/editor/lineAuthor/lineAuthorIntegration.ts
@@ -1,6 +1,11 @@
 import type { Extension } from "@codemirror/state";
-import type { EventRef, TAbstractFile, WorkspaceLeaf } from "obsidian";
-import { MarkdownView, Platform, TFile } from "obsidian";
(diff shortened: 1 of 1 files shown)


Vinzent03:
Did you test and verified the cause? The vault modify event should already be debounced as Obsidian does not store the file to disk on every keystroke.

mrueg:
Thanks for the quick response here. You're correct, it's not written to disk, I made a wrong assumption there. This one can be closed. :)

PR #1114, opened 2026-07-13
fix(auto): prevent immediate sync on startup when last backup time is missing

mrueg:
On a fresh install or when local storage is cleared, getLastAutoBackup(), getLastAutoPull(), and getLastAutoPush() return null, resulting in Invalid Date from new Date(null ?? "").

This causes the time differences computed by this.diff() to return NaN. Passing NaN to window.setTimeout() resolves to an immediate 0ms/1ms callback. This triggers unwanted and performance-degrading git operations immediately on boot instead of waiting for the configured intervals.

This change fixes the issue by checking for invalid Date states in diff() and returning the setting directly.

### DIFF
--- a/src/automaticsManager.ts
+++ b/src/automaticsManager.ts
@@ -261,6 +261,9 @@ export default class AutomaticsManager {
      * This is done by the difference between the setting and the time since the last auto action, but at least 0.
      */
     private diff(setting: number, lastAuto: Date) {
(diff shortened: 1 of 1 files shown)


mrueg:
@Vinzent03 looks like there was some temporary issue with CI, I've pushed the open PRs again. Feel free to review, if your time permits :)

Vinzent03:
Thanks!

PR #1116, opened 2026-07-13
fix(git): support recursive flag directly in rmdir for isomorphic-git

mrueg:
The custom DataAdapter's rmdir method extracted the recursive option via opts?.options?.recursive. However, isomorphic-git passes the options object directly as the second argument (following standard Node.js FS APIs), meaning that the recursive option resides directly on opts?.recursive. This caused directory cleanups during checkouts or clones to fail to delete recursively.

This change fixes options parsing by supporting both opts?.recursive and opts?.options?.recursive.

### DIFF
--- a/src/gitManager/myAdapter.ts
+++ b/src/gitManager/myAdapter.ts
@@ -111,8 +111,10 @@ export class MyAdapter {
         return this.adapter.mkdir(path);
     }
     async rmdir(path: string, opts: any) {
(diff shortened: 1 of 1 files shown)


Vinzent03:
That's a good catch. Thanks.

mrueg:
@Vinzent03 thanks for the merge! Would you be able to tag a new release for it?

Issue #579, opened 2023-08-15
Support for rename of files


BaumiCoder:
# Feature description
If I rename a note (or a file in general) in obsidian, it should be also renamed in git (`git mv`).
# Use case for the feature
I want to rename one of my notes in obsidian. If there are many other notes with links to this note, I change the name directly in obsidian to update all links to the new name. The renamed note should be also rename in git, so that the history of the file doesn't get lost.
## Actual workaround
Actual I have to do a workaround like this to achieve it:

1. Rename the note in obsidian, to update the links.
2. Rename it back to the old name outside of obsidian, with file explorer or on the shell.
3. Make an `git mv` on the note.

# Further ideas for the feature
- A option in the settings to switch it the "automatic git move" on and off could be wise. (Default ON?)
- Maybe the moving of a file from one folder to another should be also handled. I doesn't see big benefit of that, but technically it is quit similar.

Vinzent03:
You just have to stage the files. Git will identify the rename and properly mark it as a rename. 
Duplicate of #62

BaumiCoder:
Thanks for the quick response. 

I expect that a command like `git log renamedNote.md` also shows the commits to the file before the rename (like in SVN). That isn't the case. But now I looked at it in detail. That's the normal behaviour. Using `git mv` doesn't make it better.

Issue #62, opened 2021-04-04
Feature Request: Use git mv when moving or renaming notes


kalitara:
When files are moved or renamed with Obsidian, git sees them as being simply deleted and recreated--losing all revision information.

If the hooks are available, it would be nice to have this plugin use git mv when Obsidian performs a move or rename operation on a note.

Vinzent03:
That's normal git. If you run `git add .` and look again in Obsidian or run `git status` you see it detects the rename.

drozzy:
I didn't even know `git mv` existed lol. Been doing git add all my life and it detects renames.

Vinzent03:
@kalitara Can I close this? Or do you have further questions?

kalitara:
Yes. Go ahead and close. I just verified for myself that "git add ." does identify files that were renamed. I was looking at git status before running add.

Thank you.

PR #1118, opened 2026-07-14
fix: handle file renames as atomic changes

metrovoc:
## Problem

When a file is renamed in Obsidian, the plugin handles the two paths (deletion + untracked file) independently. Staging only the new path and committing leaves the old file's deletion behind, so the file ends up duplicated. Case-only renames can also be missed when `core.ignorecase=true`.

Related: #62, #579.

## Changes

- Record Obsidian's rename events to pair old and new paths. Nothing is staged automatically.
- Staging, unstaging, or discarding either path applies to both, so a rename lands in the index atomically (`R`).
- Source control shows the pair as `old → new`; file history uses `--follow`.
- Same behavior on desktop and mobile; covers cross-directory, case-only, and folder renames, and ignored destinations.
- Partially staged states stay as separate lines to reflect the real index state. Copies and renames made outside Obsidian behave as before.

Pairings are kept in local storage; chained renames collapse (`A → B → C` becomes `A → C`) and stale entries are dropped after commit, revert, or discard.

## Validation

- `pnpm run all` and `pnpm run build` pass
- 16 dependency-free unit tests, added to CI
- Manually tested normal, case-only, newly-staged, and ignored-destination renames

Note: CI fails in `pnpm/action-setup` before any repo command runs; other recent PRs hit the same failure, so it appears unrelated to this branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,6 +3,17 @@ on:
     push:
     pull_request:
 jobs:
+    unit-tests:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v6
+            - uses: pnpm/action-setup@v6
+              with:
+                  version: latest
+            - name: Install modules
+              run: pnpm install
+            - name: Run unit tests
+              run: pnpm run test
     svelte-check:
         runs-on: ubuntu-latest
         steps:
--- a/package.json
+++ b/package.json
@@ -7,11 +7,12 @@
         "dev": "node esbuild.config.mjs dev",
         "build": "node esbuild.config.mjs production",
         "release": "standard-version",
+        "test": "node --test tests/renameTracker.test.mjs",
         "lint": "eslint src",
-        "format": "prettier --check src",
+        "format": "prettier --check src tests",
         "tsc": "tsc --noEmit",
         "svelte": "svelte-check",
-        "all": "pnpm run tsc && pnpm run svelte && pnpm run format && pnpm run lint"
+        "all": "pnpm run test && pnpm run tsc && pnpm run svelte && pnpm run format && pnpm run lint"
     },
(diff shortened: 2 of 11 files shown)


Vinzent03:
I don't quite get the reason for this pr. I haven't looked into the changes, but I would prefer that the plugin works just like usual git works. Renames are only tracked on staging them. That is how software like VSCode does this as well, doesn't it?

metrovoc:
Thanks for the prompt reply. Sorry for making this pr too big before asking. The one case I couldn't solve with plain git is a case-only rename (`note.md` → `Note.md`) with `core.ignorecase=true`, the default on macOS/Windows:

```
$ mv note.md Note.md
$ git status   # clean
$ git add -A
$ git status   # still clean — the index keeps note.md
```

Unless I'm missing something, the rename never reaches git this way, so a case-sensitive device keeps the old name. Fixing it seems to require using Obsidian's rename event and updating the index explicitly on staging.

Would you be open to a much smaller PR scoped to just that?

Vinzent03:
The option for `-A` requires the file to be already in the index which is not the case for a newly created file. Use `git add .` for this. But this should not matter, because the plugin stages all files, also files not already in the index.

metrovoc:
To clarify: in my example note.md was already committed, and there's no untracked file. With core.ignorecase=true the rename is invisible to git entirely (git add . behaves the same).

```
$ mv note.md Note.md
$ git add .
$ git status      # clean
$ git ls-files    # note.md — old name kept
```

That said, I agree with keeping the plugin aligned with plain git behavior. Feel free to close if you don't think this edge case is worth handling.

Vinzent03:
Yeah I would keep it in its current form. If the filesystem is not case-sensitive and git ignores it I think the plugin should reflect how git represents it.

PR #1122, opened 2026-07-19
fix(automatics): prevent rescheduling auto runs when plugin is unload…

mrueg:
In Obsidian, plugins can be disabled or updated dynamically. When that happens, Obsidian unloads the plugin, and obsidian-git sets gitReady = false and clears its current timers.

However, if an automatic sync (commit, pull, or push) was already executing when the plugin was unloaded, the running promise couldn't be cancelled. When the Git command finished, its completion callback would run and call the scheduling function (startAutoCommitAndSync, startAutoPull, or startAutoPush) to schedule the next run.

Because the scheduling functions didn't check if the plugin was still loaded, they would register a new window.setTimeout on the global window. When that timer fired, it would run the Git operations and register another timer, indefinitely looping in the background on an unloaded plugin.

### DIFF
--- a/src/automaticsManager.ts
+++ b/src/automaticsManager.ts
@@ -120,7 +120,17 @@ export default class AutomaticsManager {
         }
     }
 
+    private get isAutomaticsDisabled(): boolean {
+        return (
+            !this.plugin.gitReady ||
+            this.plugin.localStorage.getPausedAutomatics()
+        );
+    }
+
     private startAutoCommitAndSync(minutes?: number) {
+        if (this.isAutomaticsDisabled) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


src/automaticsManager.ts
```diff
@@ -121,6 +121,12 @@ export default class AutomaticsManager {
     }
 
     private startAutoCommitAndSync(minutes?: number) {
+        if (
+            !this.plugin.gitReady ||
```
Vinzent03, line 125: I think this repeating two lines can be added to a dedicated getter to reduce code duplication and easier extension in the future.

Vinzent03 (changes requested):

PR #1125, opened 2026-07-19
fix(editor): safeguard event unregistrations against undefined values

mrueg:
# Description
This PR safeguards the event handler unregistrations in the editor features (`SignsFeature` and `LineAuthoringFeature`) against `undefined` values during feature deactivation / plugin unloading, and removes a dead field.

---

## The Issue
When features are deactivated (or when the plugin is unloaded), `EditorIntegration` calls `deactivateFeature()` on each feature. If a feature was never activated (for example, if the feature is disabled in settings), its event variables (e.g. `workspaceLeafChangeEvent`, `fileRenameEvent`, `fileOpenEvent`, etc.) remain `undefined`.

Calling `this.plg.app.workspace.offref(this.workspaceLeafChangeEvent!)` directly with `undefined` on deactivation is unsafe. 

Additionally:
- `SignsFeature` declared and unregistered `gutterContextMenuEvent`, but it was never registered or used.

---

## The Fix
1. **Safety Checks**: Updated `destroyEventHandlers()` in both features to verify that event references exist before calling `offref()` on them:
   ```typescript
   if (this.workspaceLeafChangeEvent) {
       this.plg.app.workspace.offref(this.workspaceLeafChangeEvent);
       this.workspaceLeafChangeEvent = undefined;
   }
   ```
2. **Dead Code Cleanup**: Removed the unused `gutterContextMenuEvent` declaration and unregistration from `SignsFeature`.

---

## Verification
* Verified that compiler checks (`npm run tsc`) pass cleanly.
* Verified that formatting and linter checks (`npm run lint`) pass with no warnings.

### DIFF
--- a/src/editor/lineAuthor/lineAuthorIntegration.ts
+++ b/src/editor/lineAuthor/lineAuthorIntegration.ts
@@ -150,13 +150,34 @@ export class LineAuthoringFeature {
     }
 
     private destroyEventHandlers() {
-        this.plg.app.workspace.offref(this.gutterContextMenuEvent!);
-        this.plg.app.workspace.offref(this.refreshOnCssChangeEvent!);
-        this.plg.app.workspace.offref(this.fileOpenEvent!);
-        this.plg.app.workspace.offref(this.workspaceLeafChangeEvent!);
-        this.plg.app.vault.offref(this.fileModificationEvent!);
-        this.plg.app.workspace.offref(this.headChangeEvent!);
-        this.plg.app.vault.offref(this.fileRenameEvent!);
+        if (this.gutterContextMenuEvent) {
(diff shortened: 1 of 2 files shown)


Vinzent03:
The offref function actively checks whether the values exists or not so I don't think it is worth adding all those checks.
```
 e.prototype.offref = function(e) {
                if (e) {
                    var t = e.name
                      , n = this._
                      , i = n[t]
                      , r = [];
                    if (i && e)
                        for (var o = 0; o < i.length; o++)
                            i[o] !== e && r.push(i[o]);
                    r.length ? n[t] = r : delete n[t]
                }
```
I've implemented the gutter removement now here: 28de1e6

PR #1169, opened 2026-09-01
fix(mobile): normalize path before vault lookup to avoid File already exists (#1164)

### LINKED ISSUE #1164: [Bug]: Every checkout on mobile that modifies existing files fails with "File already exists." since 2.39.0
### Describe the bug

On Android (and iPhone, probably) with isomorphic-git backend, with the vault root being the repo root (Settings -> Git -> "Base path" empty, which is the default), every git operation that needs to update an existing file in the working directory fails with "File already exists.", surfaced as a `MultipleGitError`.

Affected operations:
- "Pull", whenever an incoming commit modifies an existing tracked file
- "CAUTION: Discard all changes"
- anything else going through isomorphic-git's checkout

Pulls that only add or delete files still work.

```shell
(Captured through USB debug with Claude Code)

MultipleGitError: There are multiple errors that were thrown by the method. Please refer to the "errors" property to see more
    at checkout (isomorphic-git, thrown after Promise.allSettled over the create/update ops)

errors[0..n]: Error: File already exists.
    at t.<anonymous> (http://localhost/app.js:1:1316703)   <- Obsidian Vault.createBinary

console: [isomorphic-git checkout] task rejected: Error: File already exists.
```

### Steps to reproduce

Observed on my Android device. Vault is the git repo root.

1. On another machine, push a commit that modifies an existing tracked file.
2. On the Android run "Git: Pull" -> `MultipleGitError`; afterwards refs point at the new commit but the modified file still has its old content and is listed as a local change.
3. Edit any tracked file on the phone, then run "Git: CAUTION: Discard all changes" -> `MultipleGitError`, nothing　discarded.

### Expected Behavior

Existing files are updated via Vault.modify/modifyBinary and each operation completes as they did in the former versions.

### Addition context

(Verified by modifying the plugin on the affected device by Claude Code.)

#### Root cause

1. isomorphic-git's checkout builds workdir paths with a template string:
   `${dir}/${filepath}`. The plugin passes dir = settings.basePath = "" for
   vault-root repos, so MyAdapter receives paths with a leading slash, e.g. "/folder/note.md".
2. Vault.getAbstractFileByPath() does not normalize its argument, so it
   returns null for these paths even though the file exists.
3. Up to 2.38.6 that null fell through to DataAdapter.write/writeBinary,
   which normalizes the path and overwrites, so everything worked.
   Commit 1d1a338 changed the fallback for non-hidden paths to Vault.create/createBinary,
   which normalizes for its existence check and therefore throws "File already exists." for every existing file.

Consequently in checkout, "create" operations succeed (file genuinely absent),
"delete" operations succeed (adapter.remove fallback normalizes), but every "update" operation throws an error.

#### Suggested fix

Normalize the incoming path (normalizePath(), or at least strip the leading "/") at the
entry of MyAdapter's methods (readFile/writeFile/stat/unlink/mkdir/rmdir/ readdir).

### Operating system

Android

### Installation Method

None

### Plugin version

2.39.0

Tofuzhu:
### Problem
On Android/iOS mobile (isomorphic-git backend), any git operation that updates an existing tracked file fails with `MultipleGitError: File already exists`. Works fine on 2.38.6, regression introduced in 2.39.0.

Affected operations:
- Pull, when the incoming commit modifies an existing tracked file
- "CAUTION: Discard all changes"
- Anything else that goes through isomorphic-git's checkout

### Root Cause
isomorphic-git builds worktree paths as `${dir}/${fullpath}`. With an empty base path (default), `dir` is `""`, so paths arrive at `MyAdapter.writeFile` with a leading slash (e.g. `/note.md`). Obsidian's vault APIs (`getAbstractFileByPath` / `vault.create`) expect a normalized, leading-slash-free path — so `getAbstractFileByPath("/note.md")` misses the existing file, falls through to `vault.create()`, and throws "File already exists" because the file is already on disk.

### Fix
1. Normalize the path with `normalizePath()` at the `writeFile` entry before passing it to vault APIs (strips leading slash, folds slashes, NFC).
2. Add a `createOrOverwrite()` helper: when the vault metadata cache doesn't know the file as a `TFile` but it exists on disk (stale cache on mobile), overwrite in place via `adapter.write*` instead of `vault.create*` — avoiding "File already exists".
3. isomorphic-git's checkout analysis already guarantees that an "update existing file" op has no local uncommitted changes (a locally modified file yields a conflict, not an update, unless `force` was requested), so overwriting here does not clobber unsaved local work.

### Tests
- Added `tests/gitManager/myAdapter.test.ts` with 4 cases (reproduces the bug: leading-slash path + existing file → should overwrite without throwing).
- Full suite: 72 tests pass, `tsc --noEmit` passes.
- With the patch stashed, the new tests reproduce the original "File already exists" error; with the patch, 4/4 pass.

### Manual verification (suggested)
On Android with empty Base path, pull a commit that modifies an existing `.md` file → should write normally without the `MultipleGitError`. Also verify "CAUTION: Discard all changes".


Vinzent03:
Thanks for your contribution, but I solved it a bit differently in bac5123

PR #1174, opened 2026-09-09
fix: resolve submodule paths against the repo root, not the vault folder

dsvolk:
## Problem

With **Update submodules** enabled, every commit-and-sync fails when the Obsidian vault is a subfolder of the git repository (e.g. repo at `~/repo`, vault at `~/repo/vault`, submodule at `~/repo/vault/sub`):

```
Git.cwd: cannot change to non-directory ".../repo/vault/vault/sub/"
```

`SimpleGit.getSubmodulePaths` runs `git submodule foreach` with cwd = repo root (`absoluteRepoPath`, set in `setGitInstance`) but prefixes the printed relative path with the **vault** base path. The two only coincide when the vault *is* the repo root. Because `commitAll` iterates submodules before committing the main repo, the thrown error aborts the whole backup and nothing gets committed.

## Fix

Use `this.absoluteRepoPath` as the prefix, i.e. the same directory the `foreach` command runs in. For vault == repo root the value is unchanged.

`pnpm run all` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code), but I personally reviewed it :)


Vinzent03 (approved):
Thanks for the fix!

PR #1175, opened 2026-09-12
feat: merge conflict resolution UI in editor

ViscousPot:
Conflict resolution in the editor:
- per-hunk keep ours / theirs / both
- header bar to apply a choice to the whole file
- conflict list uses checkboxes and updates as files are resolved

Image

Only shows in Live Preview. Source mode still shows the raw markers for manual edits. Tried to keep the UI as basic as possible.

Note: the purple quote lines after a conflict are markdown reacting to the `>` in `>>>>>>>`. Pre-existing, and fixing it properly needs parser-level work, so I left it out.

Happy to split, rename or drop anything


Vinzent03:
I planned on improving the conflict handling for a while. This is a great starting point!
Some thoughts though about the current style. It no longer allows manually editing the conflict markers. At least for code, I often have/want to manually resolve the conflict. In the current version I would need to use the keep both button, but this merges both texts and I no longer no which text is from which. Though I'm unsure how relevant that is for a note taking scenario.

What do you think about instead of replacing the text with the custom widget to highlight the usual text with different colors for ours/theirs/base and keep the buttons? I'm not entirely sure how doable that is with codemirror though.

As a follow up I plan on removing the extra conflict note file and integrate that into the source control view. But that is out of scope here for now.

ViscousPot:
I've taken a stab at updating the UI and this iteration is definitely more "opinionated" as opposed to the previous more "default" look. Limited myself to existing colours to try and stay within theme and I really love the new sync menu style so I tried to emulate that with the new buttons.

Screenshot from 2026-09-18 00-20-08


Also, I did spend some time trying to prevent the conflict marker lines from affecting the rendering of the markdown around but it doesn't seem to be a solved problem sadly and I didn't want to introduce a bunch of jank to make a few lines render as plaintext

Screenshot from 2026-09-18 01-58-40

Happy to make any further changes as needed and am probably going to take a stab at adding merge conflict UI into source control view, in a separate PR, if that's ok :)

I'm thinking just conditionally add a merge conflict dropdown section to the top of the source control UI (below the buttons, above the other dropodowns) that has some sort of indication of if the file has been resolved (same function as the new checkboxes in this PR)

Vinzent03:
I've updated the colors now to use blue and cyan from the Obsidian color palette. Additionally, I added a very light purple color for the base text.
What do you think of this new style?

ViscousPot:
image

Love how it's looking 🙌🏾

Vinzent03:
I've iterated a bit further on the design. What do you think of this style in contrast to the current one where the only the actual text is highlighted? 

image

ViscousPot:
Ok! I love this iteration and spotting a few tiny updates I suggest (haven't made any edits in case you disagree)

- The buttons should be squared off (or at the least the top and bottom right corners of the "keep both" button) in the conflict markers. Looks weird to have those few pixels peeking out
- We should probably add a "Keep 03606df" button when we have these multi split conflicts (I'm actually not familiar with this conflict marker layout so making a few assumptions)
- Love the colours and the top + bottom marker highlight. We should probably also remove the lines for the separator markers and just extend their highlight to end of line while keeping them grey (like the top bottom markers)

Also, I was thinking on how we could solve these styling issues (like the "a" in your screenshot becoming a header) and actually I imagine it might be easier to apply CSS to make a whole note unstyled, as opposed to trying to style conflicting sections in particular to make them unstyled. I'm happy to explore this as a solution in a follow up PR if that sounds good? 🙏🏾 
(something like https://gist.github.com/jmatsushita/ff53393057982216260404abebc003bf)

Vinzent03:
You are right, thanks. Applied the review. 

I'm unsure about your live preview issue. If a file has much markdown syntax and has only a small conflict I don't think it makes sense to try disabling live preview for the whole note. I would like to make these changes minimally.

image

ViscousPot:
That's looking great!
Can we reorder buttons to be keep ours, base, theirs, both instead to match the text order? (again, not very familiar with this type of conflict so if the order isn't guaranteed then, no need)
Also, can we add an Keep all base button to the file header? (will all conflicts have a base section for a given file?)

Also, totally fair on not wanting to mess with the rendering in Live Preview. I can setup a snippet just for myself, I'm sure! The markdown within conflicting sections not rendering correctly, really irritates me more than it should :)

Lastly, I did create a PR to make a start on moving the file to the source control view, but I've left the file on mobile just because we don't have that conflicting array populated so I wasn't sure how to best go about it (and if it was even worth it)
https://github.com/Vinzent03/obsidian-git/pull/1181 (to be clear, no rush on this, at all 🙏🏾 )

I think I haven't said this yet, but I'm a massive fan of this plugin, and it's really great to see you really pushing for the plugin to be better! Honestly, very grateful to be able to contribute something useful here 🙌🏾

ViscousPot:
Oh, just tried the latest changes and it seems like the buttons are showing as a different colour on my machine because of the changes in `refactor: iterate of design` 🤔 
image

Vinzent03:
A quick note on diff3: The base text is the last text that was shared between ours and theirs. So yeath it always exists.

You are right about the buttons. Fixed that.

Thanks for the kinds words and the work here to get this aspect finally improved.

Can you debug the wrong button color on your side? It works for me with the default theme in light and dark mode. Maybe it is your Obsidian version, but I highly doubt that. I'm using 1.14.2.

ViscousPot:
I think it's the browser engine being used under the hood or something (I'm on Linux)
And `oklch` was picking up some hue from the `--interactive-normal` grey. I tried replacing that with transparent, but changing `oklch` to `oklab` looked more like your screenshot. I'm pretty sure `oklab` is what we want because it drops the hue part.

Vinzent03:
Yeah you may be right. I'm on Linux as well, but my installer version is rather old with 1.13.4. This should be good for now.

PR #1181, opened 2026-09-18
feat: merge conflict in source control

ViscousPot:
Moving the merge conflict resolution from being in a file to being in source control view as we discussed. I've actually kept the file around on mobile, so this is mostly tested desktop only. As usual my design skills aren't the best so would love your feedback on that :)

image


Vinzent03:
I've changed now quite a bit. Mobile also supports a list of conflicted files and generally internals have changed. Regarding the ui, I've made conflicted files a usual section like the others and remove the files from the usual changed section so they don't appear twice. There are currently too many notices when a merge fails on pull, but will solve that after the pr is merged.
Do you have any comments on the new state?

ViscousPot:
image

As with the other PR, your UI changes look much better than my attempt and I had a play with all the flows that I expect to work, and it all behaves great!! :)

Especially love this new dialog
image


If I were to nitpick, I think the new "merge in progress" button might make that top row of the source control view wider than previous so by default the new button appears on a second row. But I barely thought about it when I dragged to adjust the width of that sidebar and Obsidian remembers that so it should be a one time fix I imagine 👍🏾 

(Also, had a moment where the merge conflict UI wasn't updating in the source control view but it stopped happening when I looked into it so I think I scared it off 😄 I'll let you know if it reoccurs though, of course)

Vinzent03:
Great you like it. What exactly do you mean with the view not updating? That the conflicts count was incorrect or the green check was missing when all conflicts were resolved? May it just the delay of Obsidian actually storing the changes in the editor to disk?

ViscousPot:
Sorry should have been more specific. It was that after hitting sync to trigger a merge conflict, I saw the little message saying that there are merge conflicts in n files, and the files updated their content but until I reloaded the app, the source control view didn't show the merge conflict UI. I reproduced it a few times and then added logs to find the root cause but that fixed it, after which I removed the logs and it continued to work fine.
Maybe some cache issue on my end or something else, but doesn't seem like anything to worry about 👍🏾

Vinzent03:
Alright. I have an idea what the issue could be and will fix it with the notifications besides this pr.

Thanks again for starting this.

ViscousPot:
Amazing to see it merged, you're the best!

PR #1182, opened 2026-09-20
fix(simpleGit): apply textconv when fetching blob content for show()

kdudzik:
## Problem

`GitManager.show()` fetches blob content with `git show <rev>:<path>`. This command returns the raw blob and never applies a repo's `diff.*.textconv` filter, unlike `git diff`/`git log -p`, which do run it.

For a file tracked with a clean/smudge + textconv setup (e.g. [git-crypt](https://github.com/AGWA/git-crypt), git-secret, or any custom textconv driver), this means:

- **Diff view** (`splitDiffView.ts`): the "old" side calls `show()` and renders as raw ciphertext, while the "new" side (read from the working tree, already smudged) renders as plaintext.
- **In-editor change signs** (`signsProvider.ts`): same issue, since it also calls `show()` to get the comparison content.

Meanwhile the CLI (`git diff -- path`) already shows correctly decrypted content for the same repo/file, since it goes through the configured textconv driver — so this is specific to how the plugin fetches blob content, not the git-crypt config itself.

## Fix

Use `git cat-file --textconv <rev>:<path>` instead of `git show <rev>:<path>`. `cat-file --textconv` runs the same textconv driver as `git diff`, and is otherwise equivalent to `show` for plain blob retrieval. Repos without a textconv filter configured are unaffected.

## Testing

- `pnpm run tsc`, `pnpm run svelte`, `pnpm run lint`, `pnpm run format` all pass.
- `pnpm run test`: 81/82 pass; the 1 pre-existing failure (`updates submodules only when no tracking branch exists`, an unrelated log-message wording assertion) also fails on `main` before this change.
- `pnpm run build` succeeds.
- Manually verified against a local vault using git-crypt: before the fix, the diff view's old side showed the `GITCRYPT` header + ciphertext; after, it shows decrypted plaintext matching `git diff` output in the terminal.

🤖 Written with [Claude Code](https://claude.com/claude-code); I reviewed the change and test output myself.

https://claude.ai/code/session_01KNyGcqQN4B7qpNsak6ve8G


Vinzent03:
`git show` also supports the `--textconv` option so I would rather want to just add the argument to the existing `show` method from simple-git instead of switching to `git.raw`.

kdudzik:
Good point, changed

Vinzent03:
LGTM now. I also needed to update the config of simple-git to allow this.

